PR #126, opened 2026-07-05
oidc: tolerate trailing-slash issuer (Authentik) in discovery

### LINKED ISSUE #125: OIDC discovery fails with Authentik: issuer trailing slash stripped then strict-matched
## Summary

Native OIDC/SSO cannot work with **Authentik** (which is listed as a compatible provider in `docs/oidc.md`) because scrumboy force-strips the issuer's trailing slash and then requires a strict character-for-character match against the issuer advertised in the provider's discovery document. Authentik **always** advertises a trailing-slash issuer, so discovery permanently fails.

Tested with scrumboy `v3.18.23`.

## Symptom

`GET /api/auth/oidc/login` returns 503, and the server log shows:

```
oidc: discovery/login error: oidc discovery failed for "https://auth.example/application/o/scrumboy":
  oidc: issuer did not match the issuer returned by provider,
  expected "https://auth.example/application/o/scrumboy"
  got      "https://auth.example/application/o/scrumboy/"
```

## Root cause

1. `internal/config/config.go` `normalizeIssuer()` does `s = strings.TrimRight(s, "/")`, so `SCRUMBOY_OIDC_ISSUER` is always slash-stripped — the user cannot "add the slash back" (the usual workaround for this class of provider).
2. `internal/oidc/oidc.go` `ensureProvider()` calls `gooidc.NewProvider(ctx, s.cfg.IssuerCanonical)`, which strict-compares the discovery document's `issuer` to the passed value.
3. **Authentik always emits a trailing-slash issuer**, in every `issuer_mode` (`per_provider` → `…/application/o/<slug>/`, `global` → `https://auth.example/`). There is no Authentik setting to remove it. This is a well-documented Authentik interop wrinkle that affects many strict OIDC clients.

So the normalized (no-slash) issuer can never equal Authentik's (slash) issuer, and discovery always fails. `docs/oidc.md` even states "the `iss` claim must match the normalized value character-for-character," which directly contradicts the "Authentik" entry in the compatible-providers list. (The doc examples all use Keycloak, whose issuer has no trailing slash — likely the only provider actually validated.)

## Proposed fix

If strict discovery fails, retry once trusting the exact issuer the provider uses (configured issuer + `/`) via go-oidc's `InsecureIssuerURLContext`. The provider's ID-token `iss` carries the same trailing slash, so the verifier still validates correctly. Providers without a trailing slash (Keycloak, Entra, Zitadel, …) take the first branch and are unaffected.

```diff
--- a/internal/oidc/oidc.go
+++ b/internal/oidc/oidc.go
@@ func (s *Service) ensureProvider(ctx context.Context) (*gooidc.Provider, *gooidc.IDTokenVerifier, error) {
 	p, err := gooidc.NewProvider(ctx, s.cfg.IssuerCanonical)
+	if err != nil {
+		// Some IdPs (e.g. Authentik) advertise an issuer that differs from our
+		// normalized (slash-stripped) issuer only by a trailing slash, which
+		// go-oidc's strict issuer check rejects. Retry once, trusting the exact
+		// issuer the provider uses (its ID token iss carries the slash too).
+		slashCtx := gooidc.InsecureIssuerURLContext(ctx, s.cfg.IssuerCanonical+"/")
+		p, err = gooidc.NewProvider(slashCtx, s.cfg.IssuerCanonical)
+	}
 	if err != nil {
 		return nil, nil, fmt.Errorf("oidc discovery failed for %q: %w", s.cfg.IssuerCanonical, err)
 	}
```

I've built this patch on top of `v3.18.23` and verified it against a live Authentik instance: `/api/auth/oidc/login` now correctly 302-redirects into Authentik's `/application/o/authorize/` endpoint, and the full discovery succeeds. Happy to send it as a PR if you'd like.

Thanks for scrumboy!

vicmike:
Fixes #125.

As discussed in #125, native OIDC/SSO can't work with Authentik because scrumboy strips the issuer's trailing slash (`config.normalizeIssuer`) and then go-oidc strict-matches it against the issuer in the discovery document. Authentik always advertises a trailing-slash issuer in every `issuer_mode`, so discovery permanently 503s and the "Authentik" entry in `docs/oidc.md` is unusable.

**Fix:** in `ensureProvider`, if the strict `gooidc.NewProvider` discovery fails, retry once trusting the exact issuer the provider uses (configured issuer + `/`) via go-oidc's `InsecureIssuerURLContext`. The provider's ID-token `iss` claim carries the same trailing slash, so the verifier still validates correctly. Providers without a trailing slash (Keycloak, Entra, Zitadel, …) take the first branch and are unaffected. One hunk, 8 lines.

**Verification:** built on this change and run against a live Authentik instance — `/api/auth/oidc/login` now 302-redirects into Authentik's `/application/o/authorize/` endpoint and full discovery succeeds. (It's been running in a homelab deployment against Authentik for several days.)

Commit is DCO signed off per your request.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,20 @@
 
 > **Upgrades:** No breaking changes in **3.7.x** / **3.8.x** / **3.9.x** / **3.10.x** / **3.11.x** / **3.12.x** / **3.13.x** / **3.14.x** / **3.15.x** / **3.16.x** / **3.17.x** / **3.18.x** unless noted below.
 
+## [3.18.25] - 2026-07-05
+
+### Fixed
+
+- **OIDC trailing-slash issuer** - Discovery accepts the normalized configured issuer or exactly the same issuer with one trailing slash (e.g. Authentik) without disabling issuer validation; stored OIDC identities still use the canonical configured issuer.
+
+### Documentation
+
+- **OIDC** - Updated issuer matching, discovery, and troubleshooting notes for trailing-slash providers.
+
+### Tests
+
+- **OIDC** - Discovery fallback, trailing-slash login identity storage, and rejection of mismatched ID-token issuers.
+
 ## [3.18.24] - 2026-07-01
 
 ### Added
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 <p align="center">
   <img width="372" src="internal/httpapi/web/githublogo.png" alt="scrumboy logo" />
   <br />
-  <img src="https://img.shields.io/badge/version-v3.18.24-blue" alt="version" />
+  <img src="https://img.shields.io/badge/version-v3.18.25-blue" alt="version" />
   <img src="https://img.shields.io/badge/license-AGPL--v3-orange" alt="license" />
   <img src="https://img.shields.io/badge/i18n-23%20languages-yellow" alt="i18n" />
   <a href="https://github.com/markrai/scrumboy/actions/workflows/ci.yml">
--- a/docs/oidc.md
+++ b/docs/oidc.md
@@ -66,13 +66,13 @@ All token handling happens on the server. The browser never sees access tokens o
 
 **Identity**
 
-- Identity is tracked by `(issuer, subject)` — the provider's `iss` + `sub` claims.
+- Identity is tracked by `(issuer, subject)` — Scrumboy's normalized configured issuer + the token's `sub` claim.
 - Email is not used as a join key. Changing your email at the provider does not break the link; the row in Scrumboy keeps the email from first login until you change it in-app or via admin.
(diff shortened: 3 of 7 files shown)


markrai:
Awesome! I’m going to make a small maintainer follow-up on this PR so we can keep the retry path strict and preserve normal issuer validation. I’ll keep your original commit in the PR and add the regression coverage before merging.

markrai (commented):
The final version preserves normal go-oidc issuer validation by retrying only the exact trailing-slash issuer variant, avoids InsecureIssuerURLContext, keeps stored OIDC identities canonical/no-slash to avoid identity churn, and adds regression coverage for canonical issuers, trailing-slash issuers, mismatched issuers, and ID-token issuer enforcement.

PR #127, opened 2026-07-12
ci: publish multi-arch (amd64 + arm64) images

jordanfelle:
## Summary

The published image (`ghcr.io/markrai/scrumboy`) is currently **amd64-only**, so it can't run natively on arm64 hosts — Apple Silicon, AWS Graviton, Azure Ampere, Raspberry Pi, etc. This PR adds `linux/arm64` alongside `linux/amd64` so a single multi-arch manifest is published.

## Changes

**`Dockerfile`**
- Pin the build stage to `--platform=$BUILDPLATFORM` and pass `TARGETOS`/`TARGETARCH` into `go build` (`GOOS`/`GOARCH`).
- Because the build already sets `CGO_ENABLED=0`, the Go toolchain runs **natively on the runner and cross-compiles** the target binary — no emulation of the compiler, so the arm64 build is nearly as fast as amd64.

**`.github/workflows/docker-publish.yml`**
- Add `docker/setup-qemu-action@v3` — the runtime stage's `RUN mkdir -p /data` executes in the target arch, so buildx needs QEMU on the amd64 runner for that one step.
- Add `platforms: linux/amd64,linux/arm64` to `docker/build-push-action`.

## Why it's safe

- buildx emits a single **multi-arch manifest list**, so existing `amd64` pulls are unaffected — Docker just picks the matching arch. arm64 hosts get a native image instead of "no matching manifest".
- The only emulated step is a trivial `mkdir`; the expensive Go compile stays native via `$BUILDPLATFORM` cross-compilation.
- No runtime/behavior changes to the app.

## Verification

- The publish workflow already runs on `pull_request` with `push: false`, so **this PR's own CI builds both `linux/amd64` and `linux/arm64`** (without pushing) — the checks here validate that the arm64 build compiles.
- Separately, I confirmed the current amd64 image runs correctly under arm64 emulation on an Azure Ampere host; this PR is to get a *native* arm64 image so emulation is no longer needed.

DCO: the commit is signed off (`Signed-off-by`).


markrai:
Hi @jordanfelle - Thank you so much! Let me take a look

tungstendev9:
nice!

PR #128, opened 2026-07-13
Add SMTP email support for self-service password reset

jordanfelle:
## Summary

Scrumboy currently has no email-sending capability — password reset only works via an admin-generated link (Settings → Users → Password) that an admin must hand-deliver out of band. #9 mentions email-based reset as a planned feature; this PR adds it as an optional, additive feature.

- **New `internal/mailer` package**: sends plain-text email over SMTP using only the Go standard library (`net/smtp` + `crypto/tls`) — no new dependency. Supports STARTTLS (587), implicit TLS (465), and a plaintext `none` mode for local dev catchers. Fails closed if STARTTLS is required but the server doesn't advertise it (no silent plaintext fallback).
- **New `SCRUMBOY_SMTP_*` config**: `HOST`, `PORT` (default 587), `USERNAME`/`PASSWORD` (optional), `FROM`, `TLS_MODE`, `DEBUG`. Gated by an `SMTPConfigured()` check that mirrors the existing VAPID/push pattern — partial or invalid config is ignored (not an error), with a distinct startup log line for each state (`smtp: enabled`, `smtp: disabled`, `smtp: partial or invalid config ignored`).
- **Async delivery**: a bounded in-memory queue + worker with up to 3 attempts and backoff for **transient** failures, mirroring the existing webhook delivery architecture. **Permanent** failures (SMTP 5xx replies and local validation/config errors such as invalid From or STARTTLS misconfig) are logged once and not retried.
- **New `POST /api/auth/request-password-reset`**: lets a user request their own reset email. It always returns an identical generic response regardless of whether the email is registered, whether SMTP is configured, or whether `SCRUMBOY_ENCRYPTION_KEY` is set — this is intentional, to avoid account enumeration. Self-service emails require a validated `SCRUMBOY_PUBLIC_BASE_URL`; reset links are built from that value, not from request `Host`/`X-Forwarded-Proto` headers. Reuses the existing password-reset token primitive (`internal/auth/tokens`) and the existing dual-key (IP + email) rate limiter. The admin-generated reset-link flow is completely unchanged and remains available regardless of SMTP config.
- **Docs**: `docs/smtp.md` (mirroring `docs/vapid.md`'s structure), `README.md`/`FAQ.md` cross-links, `scrumboy.env.example` and `docker-compose.yml` passthrough examples.

## Test plan

- [x] `go build ./cmd/scrumboy`
- [x] `go vet ./...`
- [x] `gofmt -l` clean on all touched/added files
- [x] `go test ./... -race -count=1 -timeout=5m` — full suite green, no new races
- [x] New test coverage:
  - `internal/mailer`: STARTTLS (with/without auth), auth failure, implicit TLS, plaintext mode, dial failure, RCPT rejection, STARTTLS-required-but-unsupported (fails closed), header-injection rejection, permanent-vs-transient classification — all against a hand-rolled fake SMTP listener (`internal/mailer/mailertest`)
  - `internal/httpapi`: mail queue (ordering, capacity drop, concurrency), mail worker (retry/backoff, permanent failure single-attempt, graceful shutdown flush), `SMTPConfigured`/`SMTPPartiallyConfigured`
  - `request-password-reset` handler: enumeration-safety (byte-identical responses across existing/non-existing email, SMTP-not-configured, encryption-key-not-configured branches), rate limiting (dual-key IP+email), anonymous-mode 404, CSRF exception, malformed JSON, and an end-to-end test confirming self-service reset links use `SCRUMBOY_PUBLIC_BASE_URL` (not attacker-controlled request headers) and the token parses back to the correct user
- [x] Manual smoke test against a real running server (no Docker available in my environment, so I stood up a small standalone SMTP catcher using the same `mailertest` package instead of Mailpit): bootstrap a user → `POST /api/auth/request-password-reset` → confirmed startup log `smtp: enabled (...)` → captured the real email → completed the reset via `POST /api/auth/reset-password` with the emailed token → confirmed old password rejected (401) and new password accepted (200) → confirmed a non-existent email gets an identical response with zero emails sent → confirmed graceful shutdown doesn't hang

Every commit is signed off (DCO).


jordanfelle:
I know this is a chonky PR so please let me know if you want me to tackle this differently. My plan was to next add the ability to get notified by email on tasks with an option for that

markrai:
@jordanfelle gonna take a look today... but I have a feeling _we're in good paws_ 😉
image

jordanfelle:
That reminded me to update my 20 yr old avatar! I will knock things off the counter though

markrai:
@tungstendev9 keep this in mind during review
[password reset link poisoning](https://github.com/coollabsio/coolify/security/advisories/GHSA-cgj8-7m5q-x5gv)

jordanfelle:
Good catch — this endpoint was building the reset link from the inbound request's `Host`/`X-Forwarded-Proto` headers, same as the existing admin-generated flow, which is exactly the password-reset-link-poisoning pattern in the linked advisory. Since `request-password-reset` is unauthenticated, an attacker could submit a real user's email with a spoofed `Host` and get a valid reset link emailed to that user pointing at an attacker-controlled domain.

Pushed a fix: added an optional `SCRUMBOY_PUBLIC_BASE_URL` config. When set, it's used verbatim as the origin for both the self-service and admin-generated reset links (via a shared `resetBaseURL` helper), ignoring the request's `Host`/`X-Forwarded-Proto` entirely. If unset, it falls back to the prior request-derived behavior for backward compat, but the server now logs a startup warning when SMTP is configured without it, and README/docs/smtp.md/scrumboy.env.example all call out setting it. Added a regression test (`TestRequestPasswordReset_PublicBaseURLOverridesSpoofedHost`) that sends a spoofed `Host: attacker.evil` and asserts the delivered email's link uses the configured base URL instead.

tungstendev9:
> Pushed a fix: added an optional `SCRUMBOY_PUBLIC_BASE_URL` config. 

Nice. We can actually improve on this. Currently, when SMTP is configured, and SCRUMBOY_PUBLIC_BASE_URL is absent, the server still creates a valid token and can email a URL derived from an attacker-controlled request header. 

No real backward compatability required here, since it's a new feature

jordanfelle:
Makes sense, and agreed on skipping compat for a feature that hasn't shipped yet. Pushed a fix: the self-service request-password-reset handler now refuses to send anything when `SCRUMBOY_PUBLIC_BASE_URL` isn't set, instead of falling back to the request's Host header. The admin-generated reset-link endpoint keeps that fallback, but it's owner-authenticated and hands the link back directly rather than emailing it, so it isn't exposed to the same spoofing.

markrai:
need to take a closer look at `Server.Close()` in internal/httpapi/server.go (line numbers 519:527)

we're cancelling the mail worker but does not wait for flush.

`main` shuts down HTTP with a 10s context, then calls `srv.Close()` and returns.

Also.. `time.Sleep`  isn't context aware... so some type of completion signal / shutdown deadline.

Will take another look tomorrow.

tungstendev9:
Doesn't look like SCRUMBOY_SMTP_DEBUG is passed anywhere? (placeholder for later?)

jordanfelle:
Pushed fixes for both open comments:

**Server.Close() / mail worker flush (markrai):** `mailWorker` now exposes a `Done()` channel closed once its final shutdown flush completes. `Close(ctx)` cancels both workers, then waits on each worker's `Done()` bounded by `ctx` (logging if the deadline hits first). `main.go` passes a fresh 10s deadline. Applied the same fix to the pre-existing webhook worker since it had the identical gap, and extracted the now-shared queue/retry-worker logic into a generic `deliveryQueue`/`retryWorker` (`internal/httpapi/delivery_queue.go`, `delivery_worker.go`) so the fix only has to exist once.

**SCRUMBOY_SMTP_DEBUG not wired anywhere (tungstendev9):** not a placeholder — it was a real gap. `cfg.SMTPDebug` was parsed but never reached `Options` or `mailer.Config`. Wired it through; when enabled, `mailer.Sender.Send` logs each attempt's connection details (addr, TLS mode, whether auth is used) — never credentials, the recipient address, or the body. Covered by `TestSend_Debug_LogsAttemptWithoutCredentialsBodyOrRecipient` and an end-to-end test through the actual request-password-reset path.

While I was in there I also ran a broader pass over the rest of the diff and fixed two more things:
- `handleAuthRequestPasswordReset` was enumeration-safe in response body but not timing — the account-exists path did extra DB work before responding, making it distinguishable by latency. Now floors total handler time to 200ms.
- `mailQueue` now logs a warning at 90% capacity, before anything is actually dropped, so a stuck relay is visible before deliveries silently vanish.

All existing tests plus the new ones pass; `go vet` clean.

markrai:
@jordanfelle @tungstendev9 

v3.19.0 is up as a test build. I'm running it on my test environment now to make sure nothing pops up and overall I'm confident that it's good to merge to `main` by tomorrow morning or so.

https://github.com/markrai/scrumboy/pkgs/container/scrumboy/1030570192?tag=test-build

`docker pull ghcr.io/markrai/scrumboy:test-build`

markrai:
@jordanfelle Do you have an email relay already set up? I'm signing up for one right now... If so, please give it a shot and see if it works for you :)

tungstendev9:
@markrai  
image

I got it to work (with some friction around setup) 
Mark I would say don't hand it off yet. Let's work on the documentation a bit so that it is crystal clear what is required as its a gated feature behind the env vars.

jordanfelle:
@markrai Yep, tested this end-to-end just now against a real relay (Resend's SMTP relay, \`smtp.resend.com:587\`, STARTTLS) — pulled this branch, built it, ran it locally with a fresh account:

1. \`POST /api/auth/bootstrap\` → created an account
2. \`POST /api/auth/request-password-reset\` → generic 200 response as expected
3. Email actually arrived in my inbox within a few seconds — subject "Reset your Scrumboy password", valid-looking token in the reset link, matches Resend's own send log (status: `delivered`)

Full round trip works as documented. One thing I ran into that's worth calling out for `docs/smtp.md` or the README: Resend's newer **domain-restricted** API keys (the ones scoped to a specific domain via their dashboard) get rejected by their SMTP relay with `535 Authentication credentials invalid` — only their unrestricted (account-wide) sending keys work over SMTP. Not a Scrumboy bug, just a Resend quirk that cost me a retry — might save the next person testing against Resend some time to note it.

jordanfelle:
@tungstendev9 @markrai Agreed on holding off — I hit setup friction too even after knowing the codebase. Here's what I did and what I'd suggest for docs.

**What I did:** pulled this branch, built it, ran it locally against Resend's real SMTP relay (`smtp.resend.com:587`, STARTTLS). Bootstrapped an account, called `POST /api/auth/request-password-reset`, got the generic 200, and the email actually arrived in my inbox a few seconds later — confirmed both via Resend's own send log (`status: delivered`) and by receiving it myself. Full round trip works correctly.

**Where the friction was**, concretely:

1. **The four required settings are scattered across several paragraphs of `docs/smtp.md` instead of being in one checklist.** You need to piece together that `SCRUMBOY_SMTP_HOST`, `SCRUMBOY_SMTP_FROM`, `SCRUMBOY_ENCRYPTION_KEY`, *and* `SCRUMBOY_PUBLIC_BASE_URL` **all** have to be set and valid before anything works — miss one and the feature silently stays disabled with no single "here's what's missing" signal beyond the startup log line (which only covers the SMTP-specific settings, not encryption key or base URL). Suggest a "Minimum required env vars" table right at the top of the doc, before the behavioral detail.
2. **No "how do I know it's actually working" section.** I ended up manually hitting `GET /api/auth/status` and grepping the startup log for `smtp: enabled` to sanity-check config before even trying the real send. Worth a short "Verifying your setup" bullet list: check `selfServicePasswordResetEnabled` in the status response, check the startup log line, then send yourself a real test reset once both look right.
3. **Provider-specific gotcha (mine — may be what you hit too):** Resend has two kinds of API keys — an unrestricted (account-wide) sending key, and a newer key type restricted to a specific verified domain via their dashboard. The domain-restricted kind gets flatly rejected by Resend's SMTP relay with `535 Authentication credentials invalid` — only the unrestricted key works over SMTP. Cost me a full retry cycle to figure out. Worth a short "Tested providers" section (Resend, and whatever else gets tried) with exactly this kind of gotcha called out, since `SCRUMBOY_SMTP_USERNAME`/`PASSWORD` are generic and don't hint at provider-specific quirks like this.

Happy to draft the doc changes myself if that's useful — just say the word.

markrai:
@jordanfelle Oh that would be really helpful! and the "tested providers" section would be useful too. (I tested using SMTP2Go using a username/pass) But we should note that in the case of an API, they would plug that into the password field, and "resend" (or whatever the provider requires) as the username.

In the meantime, I'll add that toast after the reset to acknowledge the password reset, but looks like we're close to done :)

jordanfelle:
Pushed the doc changes we talked about (043dba4):

- **Minimum required env vars** checklist up top — the four settings that all have to be valid together (`SCRUMBOY_SMTP_HOST`, `SCRUMBOY_SMTP_FROM`, `SCRUMBOY_ENCRYPTION_KEY`, `SCRUMBOY_PUBLIC_BASE_URL`), each with what you'll see in the startup log/status response if it's missing.
- **Tested providers** section: Resend (with the domain-restricted-API-key gotcha from before) and SMTP2Go (@markrai's username/password setup).
- A short note that for API-key providers the key goes in `SCRUMBOY_SMTP_PASSWORD` and the *username* is a literal the provider's docs specify (e.g. `resend`), not your account login — that's the exact confusion that would trip someone up on first setup.

Also re-ran the full round trip against `test-build` (not just my earlier local build) to make sure nothing regressed: pulled `ghcr.io/markrai/scrumboy:test-build` onto our Azure host, ran it standalone against the same Resend relay credentials our plansbot service uses, bootstrapped an account, requested a reset, confirmed delivery via Resend's own send log (`status: delivered`), received the email, and completed `reset-password` + logged in with the new password. Clean round trip, no issues. Container's torn down again after.

PR #130, opened 2026-07-13
Add OAuth 2.1 authorization server support for MCP clients

jordanfelle:
## Summary
- Adds a minimal OAuth 2.1 authorization server (RFC 8414/9728 discovery, RFC 7591 Dynamic Client Registration, PKCE-only auth-code flow, refresh rotation, RFC 7009 revocation) so MCP clients that support automatic OAuth (e.g. Claude Code's `claude mcp add --transport http`) can connect without manually minting a static API token.
- OAuth-issued access tokens plug into the existing `resolveRequestAuth` MCP auth boundary additively — static Bearer API tokens and session-cookie auth are unchanged.
- New migration `055_add_oauth_authorization_server.sql` (`oauth_clients`, `oauth_auth_codes`, `oauth_access_tokens`, `oauth_refresh_tokens`), new `internal/oauth` (PKCE + error vocab), new `internal/store/oauth.go`, new `internal/httpapi/routing_oauth.go` (discovery/register/authorize/token/revoke), new `docs/oauth.md`.
- Deliberately scoped down: public clients only (no client secrets), single fixed consent scope, one redirect URI per client, no admin UI, no refresh-token reuse-detection cascade — see `docs/oauth.md`'s "Not Implemented" section for the full cut list and rationale.

## Test plan
- [x] `go build ./...` and `go vet ./...` clean
- [x] `go test ./... -count=1 -timeout=5m` — all packages pass, including extended `TestMCP_InvalidBearerDoesNotFallBackToSessionCookie`
- [x] New `internal/oauth` unit tests (PKCE S256 match/mismatch/plain-rejected, token generation)
- [x] New `internal/httpapi/routing_oauth_test.go` integration tests: full happy path (register → authorize → consent → token exchange → MCP call), PKCE mismatch, expired code, replayed code, redirect_uri mismatch, anonymous-mode 404s, refresh rotation + reuse rejection, revoked-token rejection by MCP, DCR validation
- [x] Manual end-to-end smoke test against a running local instance: DCR, login-gated consent screen, approve/deny, token exchange, MCP tool call with the OAuth access token, revoke, anonymous-mode 404s

🤖 Generated with [Claude Code](https://claude.com/claude-code)


jordanfelle:
So 2 large PRs in 1 day wasn't the intention, let me know if I can make it easier. Just wanted this to work in claude outside of just code without having to make an oauth proxy in front of it

jordanfelle:
Did a security review of the OAuth 2.1 authorization server surface and pushed fixes for what I found. The store layer (auth code/refresh token consumption, hashing, atomic single-use redemption, unconditional-success revoke) held up well — no issues there. Findings, all fixed in this push:

**Consent-screen phishing via unauthenticated DCR (the significant one).** `POST /oauth/register` is self-service by design (RFC 7591), so anyone can register a client with an arbitrary `client_name` — e.g. "Claude Code" — pointing at an attacker-controlled `redirect_uri`. `renderOAuthConsentPage` only showed the name, never the redirect destination, so a logged-in user clicking a crafted `/oauth/authorize` link had no way to notice where the resulting authorization code (and thus a fully-scoped access token) would actually be sent. The consent page now discloses the `redirect_uri` destination and warns that a name alone doesn't confirm the client's identity.

**No rate limiting on `POST /oauth/register`.** Unlike `/oauth/token`, DCR had none — free, unbounded `oauth_clients` row creation, and a zero-cost first step for the attack above. Now shares `authRateLimit` (10/min per IP), same as token/login.

**No redirect_uri format validation at registration.** Any non-empty string was accepted. Now validated as a well-formed absolute `http`/`https` URL with a host (plain-`http` loopback URIs still allowed per RFC 8252 for native/CLI clients).

**PKCE comparison wasn't constant-time.** `oauth.VerifyPKCE` used `==`; switched to `crypto/subtle.ConstantTimeCompare`. Low practical impact (the challenge isn't itself secret) but cheap to harden.

Added tests for all four (`TestOAuth_DCRRateLimited`, `TestOAuth_DCRRejectsMalformedRedirectURI`, `TestOAuth_DCRAcceptsLoopbackHTTP`, `TestOAuth_ConsentPageDisclosesRedirectDestination`) plus updated `docs/oauth.md`. Full suite passes.

markrai:
Thanks! I'm going to try and wrap up https://github.com/markrai/scrumboy/pull/128 by today and then come to this.

jordanfelle:
Went back over this with fresh eyes from the smtp-password-reset PR's recent commits, since that branch just fixed a real repo-wide bug: \`clientIP()\` unconditionally trusted \`X-Forwarded-For\`, so any client could spoof it to get a fresh rate-limit bucket per request. That completely undermined the rate limiting I'd just added to \`/oauth/register\` (and the pre-existing one on \`/oauth/token\`) — an attacker could bypass it just by sending a different XFF value each time.

That branch's fix isn't mergeable as-is here (this branch predates the SMTP PR entirely and has none of its other plumbing), so I ported just the mechanism:

- \`clientIP\` is now a `Server` method; XFF is only honored when \`SCRUMBOY_TRUST_PROXY\` is explicitly set, otherwise `RemoteAddr` (unforgeable) is used.
- \`POST /oauth/register\` now requires \`Content-Type: application/json\` strictly. RFC 7591 clients always send this, but without the check a hostile cross-origin webpage could trigger DCR as a CORS "simple request" (\`text/plain\` needs no preflight) using unwitting visitors' browsers as distinct-IP relays — which would've defeated the per-IP rate limit by spreading registration load across many real addresses instead of one.

Added tests proving both defenses actually hold: a spoofed-XFF-per-request loop still gets rate limited (`TestOAuth_DCRRateLimitIgnoresSpoofedXFFByDefault`), and a `text/plain` POST is rejected before it can even touch the limiter (`TestOAuth_DCRRejectsNonJSONContentType`). Full suite passes.

markrai:
In the middle of the merge conflict resolution - so requesting a short code freeze 🤓

jordanfelle:
I can easily handle that if you want?

markrai:
> I can easily handle that if you want?

~~Right in the midst of it... so give me a few mins to push and then we can do a proper audit~~ 
DONE!

markrai:
We need to ensure that the OAuth discovery endpoints adhere to the same way we're handling the Host header for password reset link hijacking: `internal/httpapi/routing_auth.go`

Scenario:  attacker sends a request with Host: attacker controlled domain.. The metadata endpoint returns an issuer configuration pointing to that domain, or attacker spoofs X-Forwarded-Proto: https on cleartext HTTP.

So the fix would be to enforce the Public Base URL like we did in the earlier PR.

jordanfelle:
@markrai want me to take some stabs at it? Want to make sure before you start

markrai:
> @markrai want me to take some stabs at it? Want to make sure before you start

yes please! I'm going to plan for the hunkering down that's happening tomorrow with the wildfires 😭

jordanfelle:
Fixed the Host-header/issuer spoofing issue (commit 2676ab8): `oauthIssuer` now prefers `SCRUMBOY_PUBLIC_BASE_URL` when set, mirroring `resetBaseURL`, and only falls back to the request-derived `Host`/`X-Forwarded-Proto` when it's unset. Added `TestOAuthDiscovery_PublicBaseURLOverridesSpoofedHostAndProto` and a docs note recommending the env var behind a reverse proxy.

I also had this fully re-reviewed end-to-end for anything else in the OAuth surface (PKCE, token storage/entropy, DCR abuse, consent flow, rate limiting, migration schema, mode gating). Two more real issues came out of that, both fixed in commit bb49312:

1. **Consent POST CSRF gap beyond SameSite.** `/oauth/authorize`'s consent form relied solely on `SameSite=Lax`, but "site" for SameSite purposes is the registrable domain, not this exact origin — a form auto-submitted from any sibling subdomain sharing the session cookie's `Domain` would still carry it here. Combined with unauthenticated DCR (attacker pre-registers a client pointing at their own `redirect_uri`), that let an attacker silently mint an authorization code for a logged-in victim with no interaction beyond visiting a page on a sibling subdomain. Fixed by requiring the POST's `Origin` (falling back to `Referer`) to match this server's own origin; added `TestOAuth_ConsentSubmitRejectsCrossOriginPost`.

2. **Unbounded growth of `oauth_auth_codes`/`oauth_access_tokens`/`oauth_refresh_tokens`.** Nothing ever deletes rows from these tables, only marks them consumed/revoked — refresh rotation in particular leaves one dead row per refresh for every long-lived client, forever. Added `DeleteExpiredOAuthArtifacts`, wired into the same hourly sweep as the existing expired-projects cleanup in `cmd/scrumboy/main.go`, plus `expires_at` indexes on the two token tables so the sweep is cheap. Covered by `TestDeleteExpiredOAuthArtifacts`.

Also tightened `renderOAuthErrorPage` to escape its `body` param (no live XSS today — every call site is a static string — but the one call site with real `<a>` markup made it an easy footgun for a future caller), and documented in `docs/oauth.md` that revoking a refresh token doesn't cascade to invalidate the access token issued alongside it (they aren't linked by a shared grant id in the current schema; bounded by the 1h access-token TTL).

Everything else came back clean on close reading — PKCE S256 constant-time compare, auth-code single-use enforced by a conditional `UPDATE ... WHERE consumed_at IS NULL` inside a tx (not just an in-memory check), codes consumed *before* PKCE verification (no verifier brute-force against a live code), exact `redirect_uri` match before any redirect, tokens stored only as SHA-256 hashes with plaintext returned once, FK cascades from `users`/`oauth_clients` onto all three child tables, and all six `/oauth/*` + `.well-known` handlers 404 correctly in anonymous mode.

`go build`, `go vet`, and the full test suite (including the new tests above) all pass.

jordanfelle:
@markrai oh no,  stay safe

tungstendev9:
Guys here an audit which should be helpful in tackling some of the remaining work.

The order of implementation should probably be:

1. Redirect URI first (nothing else depends on it)
2. Issuer /HTTPS ladder (will make the OIDC work easier)
3. HTML security headers helper (should also be done before OIDC)
4. OIDC work benefits from stable issuer / header helper in place.

and then we can worry about docs.

The current Audit report:

### HIGH-01 — Issuer / discovery metadata: ungated forwarded scheme and insecure HTTP advertising

- **Severity:** **Medium** (revised; **High only** for badly configured public reverse-proxy deployments)  
- **Status:** **Partially addressed** in `2676ab8`  
- **Confidence:** High  
- **Category:** Reverse-proxy / host trust / OAuth metadata integrity  
- **Affected:** `oauthIssuer`; `isSecureRequest`; `docs/oauth.md`  
- **Current behavior (tip `bb49312`):**  
  1. **When `s.publicBaseURL` is set:** canonical issuer; Host/XFP spoof ignored. Tested.  
  2. **When unset:** `http(s)://` + `r.Host` with scheme from ungated `isSecureRequest()` (`X-Forwarded-Proto` / `CF-Visitor` without `TrustProxy`). Ignores `X-Forwarded-Host` → `r.URL.Host` (**LOW-02**).  
  3. Docs lead with “no configuration” while a later section tells proxy operators to set `PUBLIC_BASE_URL`.  
  4. Consent Origin/Referer checks compare against `oauthIssuer(r)` — with `PUBLIC_BASE_URL` set, browsers must use that public origin.  
- **What is actually wrong (not “Host-derived issuer is always unsafe”):**  
  - RFC 8414 expects the metadata issuer to match the URL from which metadata was fetched; deriving from request authority is normal for direct-TLS zero-config.  
  - Real defects: ungated forwarded proto/visitor headers; advertising `http://` AS metadata for non-loopback cleartext; wrong issuer when proxies rewrite `Host`; contradictory docs.  
  - RFC 8414 / MCP authorization expect HTTPS for authorization-server endpoints (loopback/dev excepted).  
- **Recommended fix (revised — do not always-require `PUBLIC_BASE_URL`):**  
  1. Normalized `SCRUMBOY_PUBLIC_BASE_URL` when configured.  
  2. Else `r.TLS != nil` → `https://` + request host (`r.Host`).  
  3. Else when `TrustProxy` enabled: require a valid forwarded **HTTPS** indication and use the **proxy-normalized external host** (the same canonical host field middleware establishes — today `ServeHTTP` may set `r.URL.Host` from `X-Forwarded-Host`, but only under ungated `isSecureRequest`; issuer must not keep using raw `r.Host` while trusting a different scheme source). Treat forwarded scheme + host as one origin, not independent knobs.  
  4. Else when the request host is **loopback only** → `http://` + that host. Loopback means explicitly: `localhost`, IPv4 `127.0.0.0/8`, IPv6 `::1`. Do **not** treat RFC1918 / LAN addresses (`192.168.x.x`, etc.) as loopback; LAN cleartext HTTP is a separate product decision and must not be silently grouped with loopback.  
  5. Else unavailable / fail closed — discovery (and related metadata) must return a **controlled unavailable/error** response; never silently emit an empty or guessed issuer.  
  6. Qualify “no configuration” in `docs/oauth.md` (true for loopback + direct-TLS; proxies need canonical origin / TrustProxy).  
- **Implementation constraints:** Do not trust comma-separated or malformed forwarded-host values. When `TrustProxy` is on, derive scheme and host from **one** validated external-origin decision — never mix raw `r.Host` with a separately trusted forwarded scheme.  
- **Tests needed:** direct TLS issuer; TrustProxy issuer; XFP spoof with `TrustProxy=false`; insecure public HTTP rejected; loopback HTTP accepted (`localhost`, `127.0.0.2`, `[::1]`); RFC1918 HTTP not treated as loopback unless product explicitly opts in; fail-closed / controlled error when issuer unavailable; keep existing PUBLIC_BASE_URL override test.  
- **Merge recommendation:** remediation **slot 2** (block merge until secure ladder lands)

### HIGH-02 — OIDC-only / local-auth-disabled cannot complete inline consent login

- **Severity:** **Medium** (revised from High — compatibility/UX, not an authorization bypass)  
- **Confidence:** High  
- **Category:** Deployment compatibility / AuthN completeness  
- **Affected:** `renderOAuthLoginPage`; `handleAuth` login when `LocalAuthDisabled` → 404  
- **Behavior:** Consent inline login only POSTs password JSON to `/api/auth/login`. That route 404s when local auth is disabled. No OIDC button on the OAuth login HTML. Users **with** an existing session cookie can still approve.  
- **Impact:** OIDC-only installs cannot connect MCP via OAuth unless already signed into the main app. Nobody gains unauthorized access; the flow simply fails.  
- **Evidence:** Login gate + login page JS; `docs/oauth.md` “Not Implemented” lists 2FA but **not** OIDC-only.  
- **Recommended fix:** Prefer an OIDC continuation (`/api/auth/oidc/login?return_to=…` preserving the full authorize request). At minimum, remove the dead password form and show explicit “sign in via the main app / SSO first” instructions. Document in `docs/oauth.md` / FAQ.  
- **Implementation constraint:** `return_to` must remain an **internal relative Scrumboy path** (e.g. `/oauth/authorize?…`), never an arbitrary absolute URL.  
- **Tests needed:** LocalAuthDisabled + no session; LocalAuthDisabled + existing session; OIDC `return_to` round-trip if implemented; reject absolute/external `return_to`.  
- **Gate:** remediation **slot 4** — **required before 3.20.0 release**; preferably in the same PR, but **not a hard merge blocker**

### MEDIUM-01 — No inline 2FA during consent login

- **Severity:** Medium  
- **Confidence:** High  
- **Category:** AuthN / product security consistency  
- **Affected:** `renderOAuthLoginPage` JS (`requires2fa` branch L490–492); docs/oauth.md L57, L151  
- **Behavior:** 2FA users see a message to log in at the main app first. Documented. Workaround: establish session on main SPA, reopen authorize URL.  
- **Impact:** Friction; users may abandon or seek weaker flows. Not a bypass of 2FA for already-sessioned users.  
- **Recommended fix:** Inline 2FA challenge or deep-link to main login with `return_to` authorize URL.  
- **Tests needed:** 2FA user happy path via pre-session; inline login returns requires2fa messaging.  
- **Merge recommendation:** remediation **slot 7** (follow-up)

### MEDIUM-02 — Consent/login HTML missing anti-framing / no-store / CSP

- **Severity:** Medium  
- **Confidence:** Medium  
- **Category:** Browser security / clickjacking  
- **Affected:** `renderOAuthConsentPage`, `renderOAuthLoginPage`, `renderOAuthErrorPage`  
- **Behavior:** Only `Content-Type: text/html`. No framing/cache/referrer/nosniff headers. Token endpoint sets `Cache-Control: no-store`.  
- **Impact:** Approve button can be UI-redressed in a frame; pages may be cached by shared browsers.  
- **Recommended fix:** Shared helper on login/consent/error HTML:  
  `Cache-Control: no-store`, `Content-Security-Policy: frame-ancestors 'none'; base-uri 'none'`, `X-Frame-Options: DENY`, `Referrer-Policy: no-referrer`, `X-Content-Type-Options: nosniff`.  
- **Tests needed:** Header assertions on login, consent, and error pages.  
- **Merge recommendation:** remediation **slot 3** (small, isolated, pre-merge)

### MEDIUM-03 — Shared `authRateLimit` bucket (10/min/IP)

- **Severity:** Medium  
- **Confidence:** High  
- **Category:** Availability / cross-feature interaction  
- **Affected:** `server.go` L382–384 `ratelimit.New(10, time.Minute)`; OAuth register/token; login/2FA/reset-password  
- **Behavior:** DCR and token use `Allow("ip:"+clientIP, "")`. Login uses IP+email dual key but still consumes the shared limiter.  
- **Impact:** DCR flood or token hammering can starve login (and vice versa) for a NAT IP.  
- **Recommended fix:** Dedicated limiters for DCR vs token vs login, or higher DCR-specific ceiling with separate key namespace.  
- **Merge recommendation:** remediation **slot 7** (follow-up; document meanwhile)

### MEDIUM-04 — No cleanup of expired OAuth rows; no admin client revocation

- **Severity:** Medium  
- **Status:** **Partially addressed** in `bb49312` — `Store.DeleteExpiredOAuthArtifacts` + periodic call from `cmd/scrumboy/main.go`; store test added. Admin list/revoke UI still missing.  
- **Confidence:** High  
- **Category:** Operational / storage  
- **Affected:** `internal/store/oauth.go`, `cmd/scrumboy/main.go`, docs/oauth.md  
- **Behavior (tip):** Spent/expired codes and tokens are swept; registered clients still never deleted via API.  
- **Impact:** Client-row growth / abandoned phishing client identities persist; token/code table growth mitigated.  
- **Recommended remaining fix:** Admin list/revoke; optional max clients.  
- **Merge recommendation:** remediation **slot 7** (follow-up; reaper no longer blocks)

### MEDIUM-05 — Redirect URI validation too loose (remote HTTP, fragment, userinfo, multi-URI)

- **Severity:** Medium (clearest remaining standards/security miss — **first** pre-merge fix)  
- **Confidence:** High  
- **Category:** OAuth redirect validation / RFC 9700 / MCP authorization  
- **Affected:** `isValidOAuthRedirectURI`; DCR `handleOAuthRegister` (`redirect_uris[0]` only)  
- **Behavior:** Validator accepts any absolute `http`/`https` URL with a host:

```go
if u.Scheme != "http" && u.Scheme != "https" {
    return false
}
return u.Host != ""
```

  So all of these register successfully today:

  - `http://127.0.0.1:45871/callback` — OK (native loopback)  
  - `http://localhost:45871/callback` — OK  
  - `http://random-internet-host.example/callback` — **not OK** (accepted anyway)  

  Also accepts fragments and userinfo. Docs claim “plain http loopback addresses are allowed” but the code does not enforce loopback. DCR silently uses `redirect_uris[0]` and ignores additional entries instead of rejecting `len != 1`.  
- **Impact:** Non-compliant cleartext remote redirects can be registered via unauthenticated DCR; if a user approves, the auth code is delivered over HTTP to a non-loopback host. Fragments/userinfo enable phishing display tricks. Exact-match still prevents *changing* the redirect after registration, but registration itself is too permissive.  
- **Recommended fix (exact rules):**  
  - Exactly one registered redirect URI (`len(redirect_uris) == 1`; reject extras instead of silent `[0]`).  
  - Absolute URI; non-empty hostname.  
  - Reject malformed ports via parsing that **actually detects** bad ports (not merely `u.Host != ""`).  
  - No userinfo; no fragment.  
  - HTTPS allowed for ordinary hosts.  
  - HTTP allowed only for loopback: `localhost`, `127.0.0.0/8`, `::1` (not RFC1918/LAN).  
  - Preserve exact string matching at authorize and token exchange.  
- **Tests needed (examples):**  
  - Pass: `http://127.0.0.2/callback`, `http://[::1]:49152/callback`, `http://localhost/callback`, clean `https://example.com/callback`.  
  - Fail: `http://localhost.example.com/callback`, `https://user@example.com/callback`, `https://example.com/callback#fragment`, remote `http://…`, multi-`redirect_uris`, malformed port.  
- **Merge recommendation:** remediation **slot 1** (**block merge**)

### MEDIUM-06 — No scopes; OAuth token = full user MCP authority

- **Severity:** Medium (product) / Info if accepted  
- **Confidence:** High  
- **Category:** Authorization scope  
- **Affected:** consent copy L513; MCP tools use store authz as user  
- **Behavior:** Single fixed capability text; token acts as the approving user for all MCP tools that user can invoke.  
- **Impact:** Approving a malicious DCR client (if user ignores redirect_uri warning) grants broad automation power — same class as minting an API token.  
- **Recommended fix:** Document prominently; future granular scopes.  
- **Merge recommendation:** remediation **slot 7** (follow-up / product accept)

markrai:
Thanks @tungstendev9 - will take a look tomorrow morning

jordanfelle:
Pushed fixes for the three pre-merge slots from the audit above (commit 23630f5):

**MEDIUM-05 (redirect_uri validation, slot 1).** `isValidOAuthRedirectURI` now rejects userinfo and fragments, and treats plain `http` as valid only for loopback (`localhost`, `127.0.0.0/8`, `::1` — explicitly not RFC1918/LAN). DCR now requires `len(redirect_uris) == 1` and rejects anything else instead of silently using `[0]`. Covers all the pass/fail examples listed (`http://127.0.0.2/callback`, `http://[::1]:49152/callback` pass; `http://localhost.example.com/callback`, `https://user@example.com/callback`, `https://example.com/callback#fragment`, remote `http://…`, multi-URI all fail).

**HIGH-01 (issuer/HTTPS ladder, slot 2).** `oauthIssuer` is now an explicit, ordered decision instead of `PublicBaseURL` vs. ungated `isSecureRequest`: `SCRUMBOY_PUBLIC_BASE_URL` → direct TLS (`r.TLS`, unforgeable) → `SCRUMBOY_TRUST_PROXY`-gated forwarded origin (scheme+host trusted together from one validated `X-Forwarded-Proto`/`X-Forwarded-Host` pair, proto must say `https`, malformed/comma-separated host rejected) → loopback-only plain `http` → fail closed. On fail-closed, both discovery endpoints now return `503 server_error` instead of the old always-succeeds fallback. `oauthConsentOriginAllowed` also fails closed (denies) rather than comparing against a guessed origin.

**MEDIUM-02 (security headers, slot 3).** Added a shared `setOAuthHTMLHeaders` helper applied to the login, consent, and error pages: `Cache-Control: no-store`, `Content-Security-Policy: frame-ancestors 'none'; base-uri 'none'`, `X-Frame-Options: DENY`, `Referrer-Policy: no-referrer`, `X-Content-Type-Options: nosniff`.

Added tests for each: the full redirect_uri pass/fail matrix, every rung of the issuer ladder (`TestOAuthIssuer_*`), the discovery 503 fail-closed path, and header assertions on all three HTML pages. Updated `docs/oauth.md`'s issuer section and the "no configuration" claim (qualified to direct-TLS/loopback; proxies need `SCRUMBOY_PUBLIC_BASE_URL`/`SCRUMBOY_TRUST_PROXY`) per the audit's note.

Full suite (`go build`, `go vet`, `go test ./...`) passes.

Left as follow-ups per the audit's own slot-7/slot-4 triage (not merge blockers): HIGH-02 (OIDC-only inline consent login), MEDIUM-01 (inline 2FA), MEDIUM-03 (shared rate-limit bucket), MEDIUM-04's admin revoke UI, MEDIUM-06 (scopes).

markrai:
@jordanfelle, @tungstendev9 

Just an informal roadmap for the next few days, guys:

Since Jordan is more familiar with Authentik, and since my previous testing has been with Keycloak  I will re-install it, on my testing server to simulate any gotchas that users might face.

I should be done with this PR by tonight, hopefully.

After than, rather than working on more features, I want to take a breather to knock out security related items, and harden in that direction.

Then we can jump to the opt-in board notifications which I'm actually very excited about... as not everyone uses the PWA with the notifications. Thanks again for starting work on that, Jordan!

jordanfelle:
@markrai if there is anything you want me to work on let me know.

markrai:
> @markrai if there is anything you want me to work on let me know.

v3.20.0 is up as ghcr.io/markrai/scrumboy:test-build - I will setup Keycloak tomorrow morning on the test environment.

In the meantime, I was taking a look at Snyk (https://app.snyk.io/) and it would be nice to have security scan badging giving us a clean bill of health. The only annoying thing is that a lot of the items which show up, have either been mitigated, or don't necessarily apply to us:

For example, in markdown-it@14.1.1 the advisory is there, but that particular feature is disabled anyways.. but for the sake it, best to upgrade.

Same with golang.org/x/crypto@v0.25.0 where the issues are around SSH, which we don't use anyways.. but best to upgrade anyways.

So that's the next item... to see which one of those supposed vulnerabilities are quick wins for us. 

So along with testing v.3.20.0 (Your OAuth contrib.) I think a piecemeal upgrade effort to meet Snyk's scan would be the next step - while making sure we don't introduce any breaking changes :) 

cc: @tungstendev9 


Group: Go standard library
Resolution: Go 1.25.5 -> 1.25.12

Group: go-jose
Resolution: 4.0.2 -> 4.1.4

Group: Old golang-jwt/jwt
Resolution: webpush-go 1.3.0 -> 1.4.0

Group: DOMPurify
Resolution: 3.4.3 -> 3.4.11

Group: markdown-it/linkify-it
Resolution: markdown-it 14.1.1 -> 14.2.0

jordanfelle:
@markrai I have worked with snyk a lot, one thing that really can help is just using renovate from mend to help keep things upgraded for you.

PR #131, opened 2026-07-15
Add opt-in email notifications for board activity

jordanfelle:
## Summary

Builds on the SMTP infrastructure added in #128 (self-service password reset) to add opt-in email notifications for board activity. Each user chooses per-category what they get emailed about; off by default.

- **New Settings -> Customization section**: a master **Email notifications** toggle (off by default) plus five category checkboxes:
  - Card assigned to me (on by default)
  - Card created, moved, or deleted (off)
  - Sprint activity (off)
  - Project, workflow, or tag changes (off)
  - Added to a project (on by default)
- **New `internal/store/email_notify_pref.go`**: preferences stored as a JSON blob under the existing generic `user_preferences` table (key `emailNotifications`), following the same pattern as `wallpaper`/`tagColors` — no new migration.
- **New `internal/httpapi/email_notify.go`**: an `eventbus.Consumer` (`emailNotifier`) registered alongside the existing push notifier and webhook dispatcher in the fanout. Resolves recipients per event type (assignee for `todo.assigned`, project members for `board.refresh_needed`, the added user for the new `project.membership` event), checks each recipient's own preference, and reuses the existing async `mailQueue`/`mailWorker` from #128 to send — no new send path.
- **Debounced activity email**: card/sprint/project activity is coalesced per (project, category) with a 2-minute window, so a burst of mutations (e.g. reordering many cards) sends at most one email round per category instead of one per mutation. Assignment/added-to-project notifications aren't debounced since they already target a single recipient.
- **Self-notification always skipped**: no email for self-assignment or adding yourself to a project.
- **New `project.membership` event** and a best-effort `actorUserId` added to `board.refresh_needed`'s payload — both additive to the existing SSE wire format (SSE bridge/webhook dispatcher behavior unchanged).
- **New `emailNotifyAvailable` capability** on `GET /api/auth/status` (`smtpConfigured && publicBaseURL != ""`) — no `SCRUMBOY_ENCRYPTION_KEY` required, unlike password reset.
- **Docs**: `docs/notifications.md` (mirrors `docs/smtp.md`'s structure), README/FAQ cross-links, CHANGELOG entry.
- i18n: new strings added to all 24 locale catalogs (parity-verified).

A security/correctness pass (header-injection risk in notification email content, the actor-authorized `ListProjectMembers` recipient lookup, self-notification edge cases, and preference-validation soundness) found no exploitable issues; the one substantive concern raised — a possible notification storm from unthrottled activity events — is addressed by the per-(project, category) debounce above.

## Test plan

- [x] `go build ./cmd/scrumboy`
- [x] `go vet ./...`
- [x] `gofmt -l` clean on all touched/added Go files
- [x] `go test ./... -race -count=1` — full suite green, no new races
- [x] New Go test coverage: `internal/store/email_notify_pref_test.go` (parse/validate/defaults/version), `internal/store/preferences_test.go` (round-trip + rejection), `internal/httpapi/email_notify_test.go` (category mapping, self-skip for assignment/membership, recipient resolution via real project membership, opted-out/no-email recipient skips, SMTP-not-configured no-op, debounce-under-burst)
- [x] **Real end-to-end test** (`internal/httpapi/email_notify_endtoend_test.go`): drives the full stack through a real HTTP server against a real (fake, but wire-protocol-accurate) SMTP listener — bootstrap two users, create a project, add a member, opt in via the real `/api/user/preferences` endpoint, assign a card via the real `/api/board/{slug}/todos` endpoint, and confirm the assignee's email actually arrives over `net/smtp`, addressed correctly, with a non-empty subject. Same harness pattern as the existing `TestRequestPasswordReset_SMTPDebugLogsSendAttempt` from #128.
- [x] Frontend: `npm run verify:i18n` (24 catalogs, parity + bootstrap coverage), `npm run build` (dist/ regenerated), new `email-notify-preferences.test.ts` (defaults, parse, hydrate, save-when-signed-in, no-network-when-signed-out)

Every commit is signed off (DCO).


jordanfelle:
Did an independent in-depth security pass on this one too (same depth as the OAuth audit on #130), focused on the areas an opt-in notification feature typically gets wrong: preference-endpoint IDOR, notification content/header injection, recipient enumeration, send-rate abuse, and SSRF via anything attacker-influenceable.

**Preferences endpoint (IDOR).** `PUT/GET /api/user/preferences` (reused generic endpoint, `internal/httpapi/routing_me.go`) scopes strictly to `store.UserIDFromContext(ctx)` from the session — there's no target-user parameter in the request body/query for an attacker to substitute. `SetUserPreference` dispatches `emailNotifications` values through `ValidateEmailNotifyPrefJSON`/`ParseEmailNotifyPref`, which reject anything but the known boolean fields and version. No cross-user read/write path.

**Header injection via card title / project name.** `handleTodoAssigned`/`handleRefreshNeeded`/`handleMembership` interpolate `domain.Title` and `proj.Name` (both attacker-influenceable, e.g. a hostile card title) directly into the email Subject. This looked like the obvious "\r\nBcc: attacker@…" injection vector at first glance, but `internal/mailer/mailer.go`'s `Send` calls `validateHeaderValue` on `Subject`/`To` before building the message and rejects any CR/LF with a `permanentError` — confirmed by the existing `CRLFInSubject`/`CRLFInTo` table tests in `mailer_test.go` (inherited from #128, exercised by this new send path too since it's the same `mailQueue`/`mailWorker`/`Sender.Send`). `mailWorker.isPermanent = mailer.IsPermanent` means that error doesn't retry-loop either. Body isn't a header, so CRLF there is inert. No exploitable injection.

**Recipient authorization (`ListProjectMembers`).** `handleRefreshNeeded` resolves recipients via `store.ListProjectMembers(ctx, projectID, p.ActorUserID)`, which internally calls `CheckProjectRole(ctx, projectID, actorID, RoleViewer)` — so the async email path re-derives its own authorization from the *current* membership state rather than trusting the event payload. `p.ActorUserID` itself comes from `store.UserIDFromContext(ctx)` at the point the original authenticated mutation happened (`events.go`), not from client input, and the code explicitly skips sending when `ActorUserID == 0` (unknown actor ⇒ no email path at all, correctly fails closed rather than open).

**Send-side logging.** `mailDelivery.LogRef` for this feature is `"email-notify category=%s user=%d"` — no email address, subject, or body ever reaches the delivery-failure logs (`delivery_worker.go`), so a malformed/injected subject can't leak PII into logs either.

**Rate/storm control.** Per-(project, category) 2-minute debounce (`shouldDebounce`) bounds board-activity email volume during bulk mutations; assignment/added-to-project aren't debounced but are inherently single-recipient and already gated by whatever rate limits exist on the todo-assignment/membership-add endpoints themselves.

**SSRF / configurable targets.** SMTP host/port is operator-configured at deploy time (`SCRUMBOY_SMTP_*`), never per-message. `projectURL` builds from the server's own `publicBaseURL`, never from request or event data — no attacker-controlled destination anywhere in the send path.

One low-severity, non-security note: `emailNotifier.lastSent` (the debounce map) grows for the lifetime of the process, one entry per `(projectID, category)` pair ever seen, and is never pruned for deleted projects. Bounded by project count × 5 categories, not by event volume, so not a practical DoS vector — just a minor permanent-growth data structure. Not blocking.

Nothing else came back exploitable; the write-up in the PR description's own security pass matches what I found independently. `go build`, `go vet`, and `go test ./...` all pass on this branch as pulled.

markrai:
Good Morning! Coming to this right after I finish updating some of the libraries
`https://github.com/markrai/scrumboy/tree/security/dependency-upgrades`

markrai:
@jordanfelle Merge conflicts resolved since PR has become a little stale hehe 
Now to properly go through this :)

jordanfelle:
Excited to spam everyone with my todos

internal/httpapi/web/dist/i18n/locales/ru.json
markrai, line 1: Some of these translations will probably benefit from a second pass at some point

markrai (approved):
Some items to keep in mind, despite being non-blocking:

1. Detached emailNotifier handlers (go n.handle / go n.handleProjectDeleted on context.Background()) are untracked and not awaited on shutdown - same fire-and-forget pattern as push, falls in the operational debt category.

2. lastSent is an unbounded in-memory map keyed by (projectID, category, userID) with no TTL prune, so long-lived processes can accumulate debounce entries indefinitely 

3. Immediate shutdown can race a still-running deletion (or other) notifier goroutine against notificationMailQueue.Seal(), dropping a late enqueue; track under a later lifecycle-hardening todo.

tungstendev9:
Gentlemen while I do this massive review, it would be a good idea to test v3.23.0 thoroughly 

`ghcr.io/markrai/scrumboy:test-build`

to make sure everything is solid.

@markrai 
@jordanfelle

tungstendev9 (commented):
Looking good so far but we'll have to do the language support later in a patch.

tungstendev9 (approved):

PR #133, opened 2026-07-18
mcp: send RFC 9728 WWW-Authenticate challenge on unauthenticated tool calls

jordanfelle:
## Summary
- The OAuth authorization server (#130) never got its resource-server-side counterpart wired up: an unauthenticated `tools/call` over `/mcp/rpc` always returned HTTP 200, with the auth failure folded into the JSON-RPC body (`"Sign-in required for this tool"` / `"authentication required"` as `isError` content). The legacy `/mcp` REST endpoint did return 401 for these, but never set `WWW-Authenticate`.
- Per the MCP auth spec, a client only knows to kick off the OAuth flow when the resource server responds `401` with `WWW-Authenticate: Bearer resource_metadata="<oauth-protected-resource URL>"`. Without that transport-level signal, claude.ai's "Add custom connector" flow (with Individual sign-in enabled) just connects anonymously — clicking Add never prompts sign-in, it just shows the tool list as available.
- Found this while wiring up the Scrumboy connector on claude.ai: `curl -X POST .../mcp/rpc -d '{"method":"tools/call","params":{"name":"projects.list", ...}}'` with no `Authorization` header came back `200 OK` with `isError: true` in the body instead of a `401`.
- `mcp.Adapter` now takes `PublicBaseURL` (mirrors `SCRUMBOY_PUBLIC_BASE_URL`) and builds the `resource_metadata` URL from it — same fail-closed posture as the existing OAuth discovery issuer resolution (no header at all if `PublicBaseURL` isn't configured, rather than guessing an origin from the request). Both `/mcp/rpc` and legacy `/mcp` now emit a real `401` + challenge for `BearerAuthFailed` and `CodeAuthRequired` tool errors.
- Deliberately scoped down to just the missing 401/challenge — no change to the auth *resolution* logic (`resolveRequestAuth`), token issuance, or the discovery endpoints themselves.

## Test plan
- [x] `go build ./...` and `go vet ./...` clean
- [x] `go test ./...` — all packages pass, including two new tests: `TestJSONRPC_ToolsCall_ErrorMapping_AuthRequired` (now also asserts `401`) and `TestJSONRPC_ToolsCall_AuthRequired_SetsWWWAuthenticateChallenge` (asserts the exact header value)
- [x] Manual end-to-end: built and ran locally with `SCRUMBOY_PUBLIC_BASE_URL=https://scrumboy.example.com`, bootstrapped an account, then called `tools/call` → `projects.list` with no `Authorization` header:
  - Before this change: `HTTP/1.1 200 OK`, body `{"result":{"content":[{"type":"text","text":"Sign-in required for this tool"}],"isError":true}}`
  - After: `HTTP/1.1 401 Unauthorized`, `Www-Authenticate: Bearer resource_metadata="https://scrumboy.example.com/.well-known/oauth-protected-resource"`, same JSON-RPC body


markrai:
Yep! I saw this last night. But another pressing issue I saw is that an OIDC-bootstrapped user can permanently lose access if let's say the identity provider is unavailable...  After login, Scrumboy issues its own `scrumboy_session` cookie.. with a 30-day session duration. Once they log out, lose the cookie, change browser they won't be able to sign back in. Working on it now and I'll have a branch pushed shortly... and then after that, I'll jump to this.

Since I don't have a Claude account (and I can only extrapolate based on spec, you can test the flow, once finalized.

markrai:
Hey Jordan, after taking a look, I realized that fixing the 401 was the tip of the ice-berg. It ended up turning into a much larger MCP/OAuth normalization effort. The end result addresses the same problem your PR identified though, even though the implementation took a different architectural direction.

The fixes are currently under:

https://github.com/markrai/scrumboy/tree/mcp/spec-normalization

and deployed for testing at:

https://ghcr.io/markrai/scrumboy:test-build

PR #142, opened 2026-07-20
ci: replace deprecated DCO actions

markrai:


tungstendev9:
Better to use Git's own trailer parser... I can write the Bash script for you

.github/workflows/dco.yml
```diff
@@ -4,16 +4,61 @@ on:
   pull_request:
```
tungstendev9, line 4: We're better off writing a Bash script which will have zero deps

tungstendev9 (changes requested):

.github/workflows/dco.yml
markrai: Resubmitting, take a look!

tungstendev9 (approved):
approved

PR #150, opened 2026-07-20
chore: dco skips check on dependabot opened branches

tungstendev9:
this should fix the errors you're seeing on the dependabot pulls


markrai (approved):
thank you sir!

PR #151, opened 2026-07-20
initial osv-scanner workflow

markrai:
* Recursively scans the repository source (`go.mod` and `package-lock.json`) to match the local `osv-scanner scan source -r .` command.
* Runs automatically on dependency-related pull requests, pushes to the default branches, weekly, and on manual dispatch.
* Reports only newly introduced vulnerabilities on pull requests to reduce noise.
* Uploads SARIF results for GitHub Security integration.
* Fails scheduled and default-branch scans when vulnerabilities are detected (configurable).
* Uses path filters to avoid unnecessary scans on unrelated changes.

### DIFF
--- a/.github/workflows/osv-scanner.yml
+++ b/.github/workflows/osv-scanner.yml
@@ -1,0 +1,64 @@
+name: OSV-Scanner
+
+on:
+  pull_request:
+    branches:
+      - main
+      - master
+    paths:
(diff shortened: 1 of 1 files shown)


tungstendev9 (approved):
Yeah honestly I would stick with Trivy and OSV-Scanner

PR #152, opened 2026-07-20
security: update Go dependencies for OSV findings

markrai:
**Changes**
Upgrade golang.org/x/crypto from v0.31.0 to v0.54.0 to address numerous published OSV advisories.
Upgrade golang.org/x/oauth2 from v0.21.0 to v0.36.0 to resolve the remaining fixable OSV vulnerability.
Update transitive dependencies (golang.org/x/term, golang.org/x/sys) as required by the newer module versions.

Add an `osv-scanner.toml` configuration file to document and suppress GO-2026-5932, which applies only to the deprecated golang.org/x/crypto/openpgp package. Scrumboy does not import or depend on openpgp, making the advisory non-applicable to the project.

Validation
go test ./...
go vet ./...
go mod verify
go mod tidy (idempotent)
osv-scanner scan source -r .
osv-scanner scan source -r . --no-call-analysis=go

The dependency upgrades were audited against Scrumboy's actual call paths and import graph. No API compatibility issues or behavioral regressions were identified, and no affected openpgp packages are present in the application dependency graph.


tungstendev9 (approved):
good call on the toml. I was wondering how to suppress the false-positives

tungstendev9:
@markrai we should do a version bump too

PR #154, opened 2026-07-20
bring branch up to date with main

markrai:


tungstendev9 (approved):
you might want to retire Snyk at some point. I think Trivy & OpenSSF are enough

PR #155, opened 2026-07-20
Security/trivy

markrai:
adding initial Trivy config to workflow


markrai:
@tungstendev9 
We already checked the results of this in the `security/trivy` - should be good to merge.

tungstendev9:
> @tungstendev9 We already checked the results of this in the `security/trivy` - should be good to merge.

yep.. I saw that

tungstendev9 (approved):
This would be a good value add along with OSSF

PR #156, opened 2026-07-20
security: pin workflow dependencies

markrai:
- Pin all GitHub Actions to immutable commit SHAs.
- Pin Docker base images to immutable OCI image digests.
- Preserve existing workflow behavior while improving supply-chain security.
- Improve OpenSSF Scorecard Pinned-Dependencies compliance.
- Bump patch version.


tungstendev9 (approved):
oh wow this is really good 👍

PR #158, opened 2026-07-20
security: apply least-privilege workflow permissions

markrai:
adding top-level permissions to docker-publish, and trivy configs, keeping write access limited.


tungstendev9 (approved):
approved.

PR #159, opened 2026-07-20
Security/openssf

markrai:
- testing/fuzz JSON-RPC, OIDC return paths, and HTTP authorities
 - add the openssf scorecard badge
 - license cleanup
 - added CODEOWNERS file required to enforce mandatory approval rules

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,0 +1,1 @@
+* @markrai @tungstendev9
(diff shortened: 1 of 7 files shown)


tungstendev9 (approved):
It seems like the Windows scorecard suite isn't as robust as the one on Linux... where are are getting a solid 8.0 - it's hard to predict how the GH workflow will handle it. Only one way to find out!

PR #162, opened 2026-07-21
ci(release): serialize Windows Go tests to avoid store timeouts

markrai:
Replace the current Go test command with:
go test ./... -count=1 -p=1 -timeout=15m


tungstendev9 (approved):
approved.

PR #168, opened 2026-07-23
prevent release tags from silently generating latest

markrai:
flavor: latest=false prevents release tags from silently generating latest; the explicit latest rule becomes the only way that tag can be emitted. Docker documents both the automatic behavior and the enable attribute used to control each tag rule.

The previous global concurrency group meant a manual test-build could cancel an in-progress main image build, or vice versa. The revised group keeps all test-build runs together while separating them from production branch and release builds. GitHub confirms that runs sharing a concurrency group can cancel one another when cancel-in-progress is enabled.

### DIFF
--- a/.github/workflows/docker-publish.yml
+++ b/.github/workflows/docker-publish.yml
@@ -2,18 +2,17 @@ name: Build and Push to GHCR
 
 on:
   push:
-    branches: [ main, master ]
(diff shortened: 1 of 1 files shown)


tungstendev9 (approved):
I was wondering why the latest was pointing to PR 131

PR #170, opened 2026-07-23
Fix tags.updateProjectColor 404 on every durable project's tags

jordanfelle:
## Summary

`tags.updateProjectColor` (MCP tool) unconditionally 404'd on every durable/authenticated project, making it effectively unusable outside its own test suite.

- `handleTagsUpdateProjectColor` gates on `GetProjectScopedTagByID`, which requires `project_id` set **and** `user_id IS NULL` on the `tags` row.
- Per migration `019_add_board_scoped_tags.sql`, that shape (`user_id IS NULL, project_id IS NOT NULL`) is created **only** for anonymous temporary boards.
- Every tag on a durable/authenticated project is user-owned (`GetOrCreateTag` / `setTodoTags`, invoked from `todos.create`'s `tags` param) and is merely linked into the project via `project_tags` — so `GetProjectScopedTagByID` never finds it, even though `tags.listProject` correctly reports the tag as part of the project's tag set.
- The existing tests for this tool (`TestMCPTagsUpdateProjectColorSuccess` etc.) all pass because they use the `insertProjectScopedTag` test helper to fabricate a board-scoped row directly — a shape no production code path creates for a durable project.

## Fix

Widen the existence check in `handleTagsUpdateProjectColor` to the same project-tag set `tags.listProject` already exposes (via `ListTagCounts`), covering both board-scoped rows and user-owned tags linked via `project_tags`. The subsequent write is unchanged — it already calls `UpdateTagColor`, which correctly dispatches per the tag's own scope: board-scoped tags get their shared `tags.color` updated directly, and user-owned tags get this viewer's per-viewer color preference updated in `user_tag_colors` (the same store call the HTTP board API already uses for durable projects via `UpdateTagColorForProject` in `routing_board.go`).

`tags.deleteProject` is deliberately left as-is. `DeleteTag` on a user-owned tag deletes it across *every* project that user has used it in (not just the current one), and `TestMCPTagsDeleteProjectUserOwnedTagNotFound` already codifies that `tags.deleteProject` should keep 404ing on user-owned tags for that reason. `updateProjectColor` is non-destructive, so only that path was widened.

Also updated `API.md`'s description of `tags.updateProjectColor` to drop the stale "tag must be project-scoped" wording.

## Test plan

- [x] `go build ./...`
- [x] `go test ./...` (full suite passes)
- [x] Added `TestMCPTagsUpdateProjectColorUserOwnedTagSuccess`, reproducing the real-world case: create a project, create a todo with a tag via `todos.create` (the only way tags get created on a durable project), then successfully set/verify its color via `tags.updateProjectColor` and confirm it's visible via `tags.listProject`.
- [x] All pre-existing tag tests still pass unchanged, including `TestMCPTagsDeleteProjectUserOwnedTagNotFound` (confirms `deleteProject`'s intentional restriction is untouched).

### DIFF
--- a/API.md
+++ b/API.md
@@ -413,7 +413,7 @@ Activate/close enforce sprint state (e.g. planned vs active); violations return 
 | `tags.listMine` | `{}` | `data.items` (mine tags; no `count`) |
 | `tags.updateMineColor` | `tagId`, `color` (hex or `null` to clear) | `data.tag` |
 | `tags.deleteMine` | `tagId` | `data.deleted` `{ tagId }` - only if tag is in the viewer’s mine list, then store delete |
-| `tags.updateProjectColor` | `projectSlug`, `tagId`, `color` | `data.tag` - **maintainer+**; tag must be **project-scoped** in that project |
+| `tags.updateProjectColor` | `projectSlug`, `tagId`, `color` | `data.tag` - **maintainer+**; tag must be part of that project's tag set (board-scoped or a user-owned tag used/linked in that project) |
 | `tags.deleteProject` | `projectSlug`, `tagId` | `data.deleted` `{ projectSlug, tagId }` - **maintainer+**; tag must exist as a **project-scoped** tag in that project |
 
 ### Members
--- a/API.md
+++ b/API.md
@@ -413,7 +413,7 @@ Activate/close enforce sprint state (e.g. planned vs active); violations return 
 | `tags.listMine` | `{}` | `data.items` (mine tags; no `count`) |
 | `tags.updateMineColor` | `tagId`, `color` (hex or `null` to clear) | `data.tag` |
 | `tags.deleteMine` | `tagId` | `data.deleted` `{ tagId }` - only if tag is in the viewer’s mine list, then store delete |
-| `tags.updateProjectColor` | `projectSlug`, `tagId`, `color` | `data.tag` - **maintainer+**; tag must be part of that project's tag set (board-scoped or a user-owned tag used/linked in that project) |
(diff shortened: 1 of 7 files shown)


markrai:
good catch!

Edge case though:

Clearing a color (color: null) for a user-owned tag when the user has never set a custom color before. UpdateTagColor can return ErrNotFound in that case, and tags.updateMineColor already treats that as a successful no-op. It probably a good idea to mirror that behavior here and adding a small regression test.

markrai (changes requested):
good catch!

Edge case though:

Clearing a color (color: null) for a user-owned tag when the user has never set a custom color before. UpdateTagColor can return ErrNotFound in that case, and tags.updateMineColor already treats that as a successful no-op. It probably a good idea to mirror that behavior here and adding a small regression test.

tungstendev9 (commented):
We call ListTagCounts twice but I think for the bugfix that is okay. We can later do a ProjectContainsTag helper but that's for later. I'm cool with this and the edge case @markrai brought up

jordanfelle:
Good catch — fixed in 09f8ba5: `handleTagsUpdateProjectColor` now mirrors `handleTagsUpdateMineColor`'s clear-no-op handling (treats `ErrNotFound` from clearing an unset user-owned color as a successful no-op instead of a 404). Added `TestMCPTagsUpdateProjectColorUserOwnedTagClearNoOpSuccess` to cover it. Full suite still green.

markrai (approved):
Approved + Clarifications added to docs

tungstendev9 (approved):
good morning and approved

PR #171, opened 2026-07-23
Admin-configurable default email notification preferences for new users (Phase 1 of #169)

### LINKED ISSUE #169: Admin-configurable default notification preferences for new users
## Problem

Per-user email notification preferences (added in #131) are opt-in and default to a fixed set (`assigned` + `addedToProject` on, `enabled` itself off, everything else off). There's no way for a server owner/admin to set organization-wide defaults that new users start with.

On our self-hosted instance we worked around this by hand-writing a SQLite trigger directly against the DB (`AFTER INSERT ON users` → insert a preset `emailNotifications` row) so new users land with our preferred defaults instead of upstream's. That's fragile: it lives outside the app/migrations, isn't portable across DB backends, and any admin who wants this today has to reinvent it themselves against raw SQL.

## Proposed design

Scoped narrowly to `emailNotifications` (not a generic "admin default for any `user_preferences` key" mechanism — wallpaper/tagColors etc. don't have the same admin-relevance, and keeping this reviewable/small seems better than a bigger generic system right now).

1. **Server-level env var fallback** — e.g. `SCRUMBOY_DEFAULT_EMAIL_NOTIFY` holding a JSON `EmailNotifyPref` blob (same shape as stored today), parsed/validated at startup the same way SMTP config is. If unset, falls back to the current hardcoded `DefaultEmailNotifyPref()`. This covers the "set it at deploy time, no UI needed" case and is what actually creates new users' initial preference row (replacing the current call site that always uses `DefaultEmailNotifyPref()`).
2. **Admin settings UI** — an admin-only panel (Settings, gated by `system_role = admin` the same way other admin-only actions are) to view/edit the current org-wide default live, without a redeploy. Persisted server-side (e.g. a `server_settings`-style table/row, or reusing the env var's parsed value as a mutable DB-backed override) so it survives restarts and doesn't require touching `.env`.
3. **Bulk-apply action** — an explicit admin action ("apply current org defaults to all users who haven't customized their own preferences yet") rather than a silent mass-overwrite. This needs a way to distinguish "never touched" from "customized to match the default by coincidence" — likely a `customized` bit or `updated_at`/provenance marker on the `user_preferences` row (e.g. a preference row is considered untouched if it was written by the trigger/default-application path itself, not by a user's own `PUT` to the preferences endpoint). Only rows still in that untouched state get overwritten by a bulk-apply; anything a user has explicitly changed is left alone.

New users only get the org default at creation time automatically (env var or live-edited UI value, whichever is current at that moment); existing users are never silently changed except via the explicit bulk-apply action above.

## Context

Related to the SMTP (#128) and email notifications (#131) work. Happy to help design/implement further if there's appetite — this issue is meant to get alignment on the shape before a PR.

jordanfelle:
Closes #169 (Phase 1)

## Summary

Per the discussion on #169, this scopes down to the narrower shape @markrai proposed rather than the original issue's larger 3-part design (env var + admin UI + bulk-apply):

> I think best option would be to simply have DB backed admin setting, which would be cleaner/nicer... I think if we do this effort in digestible chunks, we'll come out happier.

This PR is that first digestible chunk: a **DB-backed org-wide default**, seeded onto **new users only, at creation time**. No env var (avoids the `SCRUMBOY_SMTP_FROM`-style quoting/escaping issues @markrai flagged), no bulk-apply action, no admin UI panel yet — see [Follow-ups](#follow-ups).

## What's added

- **Migration 058** — new `org_settings(key, value, updated_at)` table for org-wide admin-configurable settings (generic shape, but this PR only wires up one key: `emailNotifyDefault`).
- **`internal/store/org_settings.go`** — `GetEmailNotifyOrgDefault`/`SetEmailNotifyOrgDefault`, reusing the existing `ParseEmailNotifyPref`/`ValidateEmailNotifyPrefJSON` validation from #131. `SetEmailNotifyOrgDefault` requires `admin` or `owner` system role (`requireAdmin`).
- **`GET`/`PUT /api/admin/settings/email-notify-default`** — new admin endpoint, added alongside the existing `/api/admin/users/*` routes in `routing_admin.go`, behind the same admin-or-owner gate `handleAdmin` already enforces.
- **`CreateUser` / `CreateUserOIDC`** (non-bootstrap path only) now seed each new user's initial `user_preferences` row (key `emailNotifications`) from whatever the org default is **at that moment**, in the same transaction as the `INSERT INTO users`.

## Why creation-time seeding, not a live-applied default

This is the key design point from the issue: an org-default change must never retroactively change an existing user's preferences.

- A user created **before** an org-default change keeps whatever was in effect when *they* were created (the hardcoded fallback, if no override existed yet).
- A user created **after** an org-default change inherits the new value.
- The bootstrap owner (the very first user) is deliberately **not** seeded from an org default — there's no admin yet to have configured one — and keeps the existing lazy `GetEmailNotifyPref`/`ParseEmailNotifyPref` fallback path unchanged.

This is proven by `TestCreateUser_ExistingUsersUnaffectedByLaterOrgDefaultChange` and the end-to-end HTTP test `TestAdminEmailNotifyDefault_PutRequiresAdminOrOwnerAndSeedsNewUsers`.

## Follow-ups (intentionally out of scope here)

- **Admin settings UI panel** (the "menu where they pick what to disable/enable" from the issue) — the API this PR adds is what such a panel would call; a natural next PR.
- **Bulk-apply action** — an explicit "apply current org default to users who haven't customized their own preferences" action, distinct from this creation-time seeding.
- The "distant future" onboarding click-through @markrai mentioned.

## Test plan

- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go test ./...` (full suite passes)
- [x] New store tests: unset fallback, admin/owner-only write, invalid JSON rejection, `CreateUser` and `CreateUserOIDC` seeding from the current org default, existing users unaffected by a later change, bootstrap owner not seeded.
- [x] New HTTP tests: `GET` reports the hardcoded default with `customized: false` before any override; non-admin `PUT` is rejected; owner `PUT` updates the org default and is reflected in a subsequently created user's own `GET /api/user/preferences?key=emailNotifications`; a user created *before* the change is unaffected; invalid JSON is rejected; unauthenticated requests are rejected.
- [x] Documented in `docs/notifications.md` (new "Org-wide default for new users" section, plus the new endpoint under "HTTP endpoints").

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,16 @@
 
 > **Upgrades:** No breaking changes in **3.7.x** / **3.8.x** / **3.9.x** / **3.10.x** / **3.11.x** / **3.12.x** / **3.13.x** / **3.14.x** / **3.15.x** / **3.16.x** / **3.17.x** / **3.18.x** / **3.19.x** / **3.20.x** / **3.21.x** / **3.22.x** / **3.23.x** unless noted below. **3.22.0** has MCP/OAuth upgrade impact — see that release.
 
+## [3.23.2] - 2026-07-24
+
+### Added
+
+- **Admin-configurable default email notification preferences for new users** - Admins/owners can set an org-wide default that newly created users inherit, via `GET`/`PUT`/`DELETE /api/admin/settings/email-notify-default`. Seeding happens at creation time only and never rewrites existing users. When no override is configured, no preference row is created for new users, so an untouched instance behaves exactly as before. `DELETE` resets to the unconfigured state (`204`, idempotent). See [docs/notifications.md](docs/notifications.md).
+
+### Changed
+
+- **`user_preferences` provenance** - New `provenance` column (migration 059; `legacy` / `user` / `org_default`) records how each preference row was written. Existing rows become `legacy`; explicit user saves are tagged `user`; org-seeded rows are `org_default`. This is groundwork so a future bulk-apply can safely target only org-seeded rows and never overwrite user-customized ones. **Upgrade note:** remove any hand-rolled `AFTER INSERT ON users` preference trigger before upgrading — explicit-column triggers keep working, but positional `INSERT INTO user_preferences VALUES (...)` triggers become structurally invalid once the column is added. Running an older binary against a 059-migrated database is unsupported (forward-only upgrades).
+
 ## [3.23.1] - 2026-07-24
 
 ### Fixed
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 <p align="center">
   <img width="372" src="internal/httpapi/web/githublogo.png" alt="scrumboy logo" />
   <br />
-  <img src="https://img.shields.io/badge/version-v3.23.1-blue" alt="version" />
+  <img src="https://img.shields.io/badge/version-v3.23.2-blue" alt="version" />
   <a href="LICENSE"><img src="https://img.shields.io/badge/license-AGPL--v3-orange" alt="license" /></a>
   <img src="https://img.shields.io/badge/i18n-23%20languages-yellow" alt="i18n" />
   <a href="https://github.com/markrai/scrumboy/actions/workflows/ci.yml"><img src="https://github.com/markrai/scrumboy/actions/workflows/ci.yml/badge.svg?branch=main" alt="CI" /></a>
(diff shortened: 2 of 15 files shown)


markrai (approved):
good to go

tungstendev9 (approved):
approved

PR #172, opened 2026-07-23
Fix MCP tool names: rename dots to underscores to satisfy Claude's tool-name regex

jordanfelle:
## Summary

Claude's MCP client validates every tool name returned by `tools/list` against `^[a-zA-Z0-9_-]{1,64}$` (letters, digits, underscore, hyphen only). Scrumboy's MCP server names its tools with dots (`system.getCapabilities`, `todos.create`, `sprints.getActive`, `board.get`, etc.), which fail that regex.

Because Claude validates the whole tool array as one schema, **a single invalid tool name breaks the entire array** — this locks up tool-calling for *every* MCP server connected to that Claude session, not just Scrumboy, until Scrumboy is disconnected. This was reported as a live production bug: a user got locked out of using Claude entirely because of it.

## Fix

Renamed all 28 Scrumboy MCP tool identifiers from dot-separated to underscore-separated everywhere they function as the wire-protocol tool name (e.g. `todos.create` -> `todos_create`, `sprints.getActive` -> `sprints_getActive`, `board.get` -> `board_get`), across:

- `internal/mcp/tool_catalog.go`, `registry.go`, `adapter.go`, `system_tools.go`
- `internal/mcp/sprint_tools.go` (the dynamically-built `"sprints." + action` tool name now builds `"sprints_" + action`)
- Human-readable error message strings in `board_tools.go`, `projects_tools.go`, `todos_tools.go`, `members_tools.go` (for consistency)
- Test files: `adapter_test.go`, `jsonrpc_test.go`, `internal/httpapi/routing_oauth_test.go`, `internal/agora/agora_test.go`
- Docs: `docs/mcp.md`, `docs/oauth.md`, `docs/agoragentic.md`, `API.md`, `README.md`

**Not renamed:** `internal/httpapi/web/modules/voice/` has a separate, unrelated internal vocabulary of "voice command intents" that also happens to use dot notation (e.g. `todos.assign`, which has no corresponding MCP tool at all). Only the literal MCP tool name values that module passes to `callMcpTool`/`callTool` were renamed (`execute.ts`, `resolve.ts`, `target-resolver.ts`, `mcp-client.ts`'s `McpToolName` union, and their tests) — the intent labels themselves are untouched.

## Test plan

- [x] `go build ./...` passes
- [x] `go test ./...` passes (all packages, including `./internal/mcp/...`)
- [x] `cd internal/httpapi/web && npm run build` passes
- [x] `npm test` — all tests touched by this change pass (`modules/voice/*`); two pre-existing failing suites (`localStorage` undefined in this Node/vitest environment) reproduce identically on unmodified `main` and are unrelated to this change
- [x] Repo-wide grep confirms no remaining dotted MCP tool name literals outside the voice-intent files


jordanfelle:
Follow-up after some internal review of this change's impact — flagging something worth your call, @markrai 

**Why the rename is needed:** Claude's MCP client validates every tool name in `tools/list` against `^[a-zA-Z0-9_-]{1,64}$`. Because it validates the whole tool array as one schema, a single name with a dot in it (`todos.create`, `board.get`, etc.) breaks tool-calling for *every* MCP server connected to that Claude session, not just Scrumboy, until the user disconnects Scrumboy entirely. That's a real reported production incident, not just a lint nit — hence the rename.

**The catch:** it's a breaking change for anyone with an external MCP client that hardcodes the old dotted names via `tools/call` or the legacy `POST /mcp {"tool": "..."}` endpoint. The original version of this PR just silently dropped support for them with no migration path or changelog note, which isn't great for existing integrations.

**Stopgap I've added (pushed as a follow-up commit):**
- Dispatch-only backward-compat aliases in `internal/mcp/registry.go` — the old dotted names still work for `tools/call`/`POST /mcp`, wired to the same handlers as the new underscore names.
- They're deliberately *not* added to `implementedTools()`/`toolCatalog()`, so `tools/list` and `system_getCapabilities` only ever advertise the new underscore names — that's the whole point of the fix, so it still has to hold.
- A `docs/mcp.md` note and a new `CHANGELOG.md` "Unreleased" entry explaining the dotted names are deprecated, still callable for now, and will be removed later.

**What I'd like your call on, since it's a product/versioning decision:**
1. How long do you want to keep the dotted-name aliases around before removing them (a specific version, a time window, or just "until we see no more traffic on them")?
2. Should this ship as a minor or major version bump given the discovery-side breaking change, even with the compat shim in place?
3. Do you want a deprecation warning logged server-side whenever an old dotted name is actually invoked, so we get visibility into whether anyone's still depending on them before we pull the aliases?

Happy to adjust the shim/changelog/docs however you'd rather handle it — didn't want to presume a removal timeline or version bump on your behalf.

markrai:
I think adding regression tests covering the legacy dotted names (tools/call and legacy /mcp) to prove the compatibility layer actually works, including required-field validation through an alias and ensuring dotted names never appear in tools/list would be a good idea.

Also, please do add a catalog invariant test that asserts every advertised tool name is unique and matches the underscore-only convention. This would prevent us from accidentally reintroducing an incompatible tool name in the future.

As for your open questions, this should be a minor.. since since existing clients using dotted names still works.

I would keep the aliases indefinitely for now since it's hard to get a count on how many people are using them. No real maintenance burden as such. 

We can skip the logging.

jordanfelle:
Sounds good, working on it, with SMTP support we are fully adopting scrumboy so expect a mini flood here as I find issues

jordanfelle:
Pushed a follow-up commit addressing all of this:

- **Regression tests for the legacy dotted names** (`internal/mcp/legacy_alias_test.go`): `tools/call` and legacy `POST /mcp` both still dispatch `projects.list`/`todos.create`-style names; required-field validation runs through the alias (`todos.create` missing `title` still returns "missing required field: title"); an unknown dotted name still errors correctly on both surfaces; and `tools/list` / `system_getCapabilities` are asserted to never contain a dotted name.
- **Catalog invariant test** (`internal/mcp/tool_catalog_invariant_test.go`): asserts every name in `implementedTools()` and `toolCatalog()` is unique and matches `^[a-zA-Z0-9_-]{1,64}$`, plus a check that every `legacyToolAliases` key is a dotted name disjoint from the current canonical set. This should catch it if an incompatible name ever gets reintroduced.
- **Minor version, aliases kept indefinitely, no logging**: updated `CHANGELOG.md` and `docs/mcp.md` to say this ships as a minor (existing dotted-name callers are unaffected) and that the aliases are a permanent compatibility shim rather than on a removal timeline — dropped the "expected to be removed in a future release" language from both the docs and the `registry.go` comment. Left deprecation logging out per your call.

`go build ./...` and `go test ./...` both pass with the new tests included.

markrai:
Okay I think we are good. The overall design is still very much your implementation.

Made a few changes focused on long-term compatibility: preserving the exact behavior for instances that never configure the feature, adding preference provenance to support a future bulk-apply safely, adding a reset endpoint, and expanding the test coverage around migrations, OIDC, reset behavior, and edge cases.

this lays cleaner groundwork for the remaining phases :)

markrai (dismissed):
Good to go!

tungstendev9 (approved):
solid

markrai:
Hey @jhotmann - I hope you're doing well!

we just did a Claude compatibility fix after realizing that Claude rejects MCP servers that advertise tool names with dots, because it validates the entire tools/list response as one schema.

But we've kept full backward computability for existing integrations.

Would love for you to weigh in since you've spent some time on this + your experience specifically with Claude

jhotmann:
Don't know how I didn't notice this issue, but yeah looks good on my end with Claude and OpenCode

PR #177, opened 2026-07-24
Add workflow/column CRUD to MCP tools

jordanfelle:
## Summary
- MCP clients had no way to add, rename, recolor, or delete a project's workflow columns -- `board_get` + `todos_move` only work against columns that already exist.
- The REST API already has this (`POST`/`PATCH`/`DELETE /api/board/{slug}/workflow`) and the store layer (`AddWorkflowColumn`/`UpdateWorkflowColumn`/`DeleteWorkflowColumn`/`GetProjectWorkflow`) was already fully built, but those REST endpoints only accept the session cookie, not the `sb_` Bearer API token, so MCP clients couldn't reach them without impersonating a browser session.
- Adds `workflow_list`/`workflow_create`/`workflow_update`/`workflow_delete` as first-class MCP tools that call the existing store methods directly, bypassing the cookie-only REST auth gap. Maintainer role or higher is required for create/update/delete, matching the REST endpoints.
- No legacy dotted aliases added, per the frozen `renamedToolsWithLegacyAliases` invariant in `tool_catalog_invariant_test.go` -- these are new tools, not renames.

## Test plan
- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go test ./...` (full suite, including new `workflow_tools_test.go` covering list/create/update/delete, permission failure, invalid color, and done-column delete rejection)


markrai (approved):

tungstendev9 (approved):
approved

PR #178, opened 2026-07-24
Configurable default board(s) for newly created users

### LINKED ISSUE #175: Configurable default board(s) for newly created users
## Problem

New users (created via `CreateUser` or `CreateUserOIDC` in `internal/store/auth.go`) start with zero rows in `project_members`. There is no "everyone board," org/workspace-wide membership, or auto-invite concept anywhere in the app — every board membership is an explicit, per-project, per-user `AddProjectMember` call by a Maintainer/Owner of that specific project.

For a team that wants everyone to land in a shared onboarding/company board automatically, this means manually inviting every new hire to that board one at a time, forever, with no way to configure it once.

## Proposed design

Following the same shape as the `org_settings` pattern from #169/#171 (email notification defaults):

1. **DB-backed admin setting** — an org-level list of project IDs (or a single "default board" to start, per the "digestible chunks" precedent from #169) that new users should be auto-enrolled into. Admin/Owner-only to view/edit, same gating as the existing admin settings.
2. **Seed at creation time only** — `CreateUser`/`CreateUserOIDC` insert the corresponding `project_members` row(s) (lowest-appropriate role, e.g. viewer/member, not Maintainer) right after the user row is created, mirroring how #171 seeds `emailNotifications`. Existing users are never retroactively touched by a later change to the default.
3. Out of scope for phase 1, left as follow-ups: an admin UI panel to pick the board(s) (would slot into the same Users/admin tab as #171's panel), and any bulk-apply-to-existing-users action.

## Context

Noticed this while auditing the membership model for something unrelated. Related to the org_settings precedent set by #169/#171. Happy to help design/implement further if there's appetite — this issue is meant to get alignment on the shape before a PR.

jordanfelle:
## Summary

- Admins/owners can configure a single org-wide default board project; newly created users (`CreateUser`/`CreateUserOIDC`) are auto-enrolled as **Viewer** (lowest-appropriate role) on it, seeded in the same transaction as user creation.
- New `GET`/`PUT`/`DELETE /api/admin/settings/default-board` endpoints, gated the same as other `/api/admin/*` routes (Admin/Owner only). `PUT` body: `{ "projectId": <number> }`; the project must currently exist (validated on write).
- Follows the same `org_settings` shape as #169/#171's email-notification org default.
- Seed-at-creation-time-only semantics: existing users are never retroactively enrolled; unset/`DELETE` means no seeding at all, so an untouched instance behaves exactly as before this feature.
- Bootstrap owner is excluded (it already has implicit access via its system role).
- If the configured project is deleted after the setting was made, account creation still succeeds — seeding is silently skipped rather than failing.

Addresses the checklist in markrai/scrumboy#175 (comment https://github.com/markrai/scrumboy/issues/175#issuecomment-5073050444):
- [x] `org_settings` key + store get/set/clear for default board project ID
- [x] Validate configured project ID exists on write
- [x] Admin API: `GET`/`PUT`/`DELETE /api/admin/settings/...`
- [x] Seed `project_members` in `CreateUser`/`CreateUserOIDC` (create txn only; lowest-appropriate role)
- [x] Unset = no seed; setting changes never touch existing users
- [x] Tests (authz, create-time seed, OIDC path, invalid IDs, reset, no retroactive enroll)
- [x] Docs + changelog

Out of scope for phase 1 (per the issue): an admin UI panel to pick the board, and any bulk-apply-to-existing-users action.

## Test plan

- [x] `go test ./... -count=1 -timeout=5m` — all packages pass
- [x] New store-level tests in `internal/store/default_board_test.go` (unset/customized, authz, missing-project validation, create-time seeding for both `CreateUser` and `CreateUserOIDC`, no retroactive enrollment, bootstrap-owner exclusion, deleted-project-skips-seeding)
- [x] New HTTP-level tests in `internal/httpapi/admin_default_board_test.go` (GET/PUT/DELETE authz matrix, seeding via the real HTTP surface, reset behavior)
- [ ] Frontend `npm test` — pre-existing `localStorage is undefined` failures reproduce identically on an unmodified checkout of this repo (verified before touching any files), unrelated to this change; not introduced or worsened here.

Closes #175


markrai (dismissed):
So I preserved the project auth boundaries. i.e. kept the Admin/Owner system-role requirement... but also required that the user has to be a Maintainer of the project.

This way, the person can't simply give the viewing rights to a newly created user, and then it turns out that they control that new user, and essentially end up now with access to a project which they never had access to in the first place.

Then I also made sure that this applies only to Durable (non Temporary/non Anonymous) projects.

tungstendev9 (dismissed):
approved

tungstendev9 (approved):
eh? I thought this was already active

PR #185, opened 2026-07-26
align MCP tags_updateProjectColor tagId path with REST temporary-boar…

markrai:
- Detects temporary boards with pc.Project.ExpiresAt != nil.
- Skips the impossible RoleMaintainer check for those boards.
- Calls UpdateTagColorForTemporaryBoard, matching the REST route.
- Keeps the Maintainer+ requirement and project-aware durable method for durable projects.
- Leaves anonymous MCP mode unavailable, as intended.

The store method also revalidates scope: board-scoped tags must belong to that project, while user-owned tags must actually be used by a todo on that temporary board. Expired boards are rejected while building the project context.


tungstendev9 (approved):
solid!

PR #186, opened 2026-07-26
Fix orphaned viewer tag color preference on backing-row rotation

### LINKED ISSUE #173: Tags are scoped per-creator (user_id, name), causing duplicate project tags with no merge path
## Summary

Tags are stored uniquely by `(user_id, name)`, not `(project_id, name)`. Two members of the same project can each create a tag named e.g. `bug`, resulting in two distinct DB rows tied to the same project. This is intentional (confirmed via `TestUserOwnedTags_UsersCanOnlyAttachOwnTags`, which asserts exactly this outcome), but it has a real, unaddressed consequence: **there is no way to consolidate duplicate same-named tags within a project**, and the duplication is regenerated automatically whenever a user without their own copy of a project tag applies that tag name to a card.

## Reproduction (via MCP tool surface, but reachable through normal UI flows)

1. Project has a tag `urgent` created by user A (tagId 3, color `#692A80`, used on 28 cards).
2. User B independently applies a tag named `urgent` to a card. Since lookup is `SELECT id FROM tags WHERE user_id = ? AND name = ?` (no `project_id` in the WHERE clause — see `internal/store/tags.go` in `setTodoTags`), this creates a **second** row (tagId 11) owned by user B, with no color.
3. Project's tag list (`tags.listProject` / any admin-facing, tag_id-scoped view) now shows two entries named `urgent` — one deletable by A, one deletable by B, with no merge operation available to either.
4. Board/card filtering (`board.get` tag filter) matches by `name` across all owners, so end users don't notice — filtering "urgent" returns cards tagged with either row. The duplication is only visible/actionable from the tag-admin surface.
5. Attempting to consolidate: deleting your own copy and reapplying the tag name to your cards does **not** attach to the other owner's existing row — it silently creates yet another new user-owned row (confirmed: color is lost again, a third distinct tagId is created).

## Root cause

- Migration `019_add_board_scoped_tags.sql` gives user-owned tags a unique index on `(user_id, name)` only; there's no `(project_id, name)` uniqueness for authenticated-project tags (only anonymous/board-scoped tags get that).
- `setTodoTags`'s user-owned branch resolves/creates via `SELECT id FROM tags WHERE user_id = ? AND name = ?`, with no awareness of an existing project tag of the same name owned by someone else.
- The one existing dedup migration (`041_deduplicate_canonical_tags.sql`) only merges duplicates from name-normalization drift *within a single owner* — never across owners in the same project.

## Why this matters

The per-user tag model (personal tag color/library) is a reasonable design choice on its own, but combined with:
- no cross-owner reuse on write, and
- no merge/consolidation tool anywhere in the product,

...duplicate project tags accumulate over time with no way to clean them up, and any admin/maintainer trying to manage a project's tag set via the tag-scoped tools (list/delete/recolor by project) will see a confusing, ever-growing set of "the same" tag under different IDs.

## Proposed direction (open for team input)

Keep per-user customization, but make the *tag identity* project-scoped rather than user-scoped:
- A tag name should resolve to a single canonical row per project, regardless of who applies it first.
- Per-user color preference stays local (this already exists as a separate concept — `user_tag_colors` — so the plumbing for "shared identity + local color" is already partially there).
- Add a `tags.merge(fromTagId, toTagId)` (or equivalent) operation so existing accidental duplicates can be consolidated by project members/admins.

Filing this for team discussion rather than proposing a specific implementation, since it touches the tag ownership model directly — happy to help implement whichever direction the team decides on.

jordanfelle:
Closes #173 (review comment 2, https://github.com/markrai/scrumboy/issues/173#issuecomment-5082087432).

## Problem

`SetViewerTagColorByName` writes a viewer's color preference onto whichever personal tag rows back a canonical name *at write time* (via `personalTagRowsForName`). The grouped listing (`listTagCountsGrouped`) only resolved that preference from rows currently used by a todo in the project (via a `LEFT JOIN user_tag_colors` scoped to the "used in project" query).

If every backing row for a name later stopped being used (e.g. the tag was removed from every todo carrying it) while a different member's row for the same canonical name became the sole backing row, the viewer's stored preference became unreachable — `pickGroupedTagColor` had nothing to fall back to, and the color silently reverted to default with no user action. This was flagged as a probably-intentional tradeoff in the original review, but it's cheap to fix properly instead.

## Fix

Added `viewerPersonalTagColorPrefs`, which resolves a viewer's color preference for a canonical name across **all** of their personal tag rows, independent of whether a given row is currently used in the project being listed. `listTagCountsGrouped` now calls it once per listing and looks the result up per group, replacing the previous per-row `viewerColor` join. The existing precedence (viewer-owned row wins, then lowest backing `tag_id`) is preserved.

## Test plan

- [x] Added `TestProjectLabels_ViewerColorSurvivesBackingRowRotation`, reproducing the exact scenario: set a color, untag the only backing row so it drops out of the listing, have another member create a new row for the same name, confirm the color still resolves.
- [x] `go build ./...`
- [x] `go test ./...` (all packages green)
- [x] All existing `TestProjectLabels_*` tests still pass unchanged, including the color-precedence tests.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


tungstendev9 (approved):
approved

markrai (approved):
already reviewed.

PR #189, opened 2026-07-26
Add assignee filter to board queries and MCP board_get

jordanfelle:
## Summary
- The board API/MCP tools support filtering by tag, search text, and sprint, but not by assignee — no way to scope a view to "assigned to me" or "unassigned".
- Adds an `assignee` filter (a user ID, or `"unassigned"`) threaded through the same plumbing as the existing tag/search/sprint filters: `GetBoard`, `GetBoardPaged`, and the per-lane list/count queries in `internal/store/board.go`.
- Exposed via `GET /api/board/{slug}?assignee=...`, `GET /api/board/{slug}/lanes/{status}?assignee=...`, the legacy `/api/projects/{id}/board` route, and the `board_get` MCP tool (new `assignee` input field).

## Not included
Web UI wiring for the filter (a picker in the board view) is left for a follow-up PR — it touches router route-parsing and several render functions, and needs a placement decision (dropdown vs. a third chip row) that's easier to validate visually than to guess blind.

## Test plan
- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go test ./... -count=1 -timeout=5m` (all packages pass)
- [x] New tests: `TestGetBoard_AssigneeFilter`, `TestListTodosForBoardLane_AssigneeFilter` covering specific-assignee, unassigned, and no-filter cases


markrai (approved):
ship it!

tungstendev9 (approved):
approved

PR #190, opened 2026-07-26
Refactor CodeQL workflow configuration

markrai:
Updated CodeQL workflow to use specific action versions and removed comments for clarity.


tungstendev9:
why are we changing this?

markrai:
> why are we changing this?

It was running in the wrong order, and the CodeQL scans weren't getting counted on each commit by OpenSSF.

tungstendev9 (approved):
approved.

PR #191, opened 2026-07-26
Fix board columns collapsing loaded cards on realtime refresh

jordanfelle:
## Summary
- `getRequestedBoardLimitPerLane()` only preserved the on-screen "Load more" count across a board refresh when a tag/search/sprint filter was active; the plain unfiltered board always requested `limitPerLane=20`
- Any realtime refresh (another user moving/editing/creating a card) then re-fetched only the first page, collapsing columns the user had expanded via "Load more"
- Fix: always preserve the current on-screen lane size (DOM-based), filtered or not — the existing preservation logic used for the filtered case now applies universally

## Test plan
- [x] `tsc --noEmit` passes
- [x] Ran `/security-review` — no high-confidence findings (no new injection/authZ/data-exposure surface; `limitPerLane` had no server-side upper bound before or after this change)
- [ ] Manually verify: expand a column past 20 cards via "Load more", trigger a board change from another session, confirm the expanded cards remain visible


markrai:
Right now the client preserves the expanded lane size across all board loads. That means if I expand a lane on Board A (say to 200 cards) and then navigate to Board B, the first request for Board B also asks for 200 cards per lane.

The expanded count should only be preserved when refreshing or changing filters on the same board. Initial loads and cross-board navigation should fall back to the default of 20. A good approach would be to preserve the DOM-derived count only when the currently loaded board is the same board being requested.

getBoard()?.project?.slug should work well here, since the router updates getSlug() before loading the new board but doesn't replace the current board object until the response arrives.

The regression test should cover:

Unfiltered same-board refresh retains more than 20 cards.
Filtered same-board refresh retains more than 20 cards.
Navigation to a different board resets the requested limit to 20.
Initial load with no existing board/DOM defaults to 20.

jordanfelle:
Pushed a fix: `getRequestedBoardLimitPerLane()` now only uses the DOM-derived lane size when `getBoard()?.project?.slug === getSlug()`, falling back to 20 otherwise. Since `getSlug()` is updated by the router ahead of the request while `getBoard()` still holds the previously loaded board until the response lands, this distinguishes a same-board refresh (filtered or not) from cross-board navigation.

Added `board-limit-per-lane.test.ts` covering the four cases you listed: unfiltered same-board refresh, filtered same-board refresh, cross-board navigation, and initial load with no board/DOM.

markrai (approved):
already reviewed

tungstendev9 (approved):
approved.

PR #192, opened 2026-07-27
Add assignee filter dropdown to board view

jordanfelle:
## Summary
- Follow-up to #189, which added `assignee` query-param support to the board API/MCP tools but explicitly deferred UI wiring.
- Adds an assignee + sort filter, exposed as an expandable panel on the search input (chevron toggle) rather than a separate topbar control — avoids cluttering the topbar on narrow/PWA widths where it already wraps. Options: assignee (All / Unassigned / Assigned to me / each board member) and sort (Default order / Newest first / Oldest first).
- Sort is threaded through the full stack, not just the UI: `store.SortOrder` (manual/newest/oldest), `ORDER BY` selection in the board query and cursor-based lane "load more" pagination (cursor switches from `rank:id` to `createdAt:id` when a sort is active), the `sort` HTTP query param, and the MCP `board_get` tool input.
- Active-filter indicator: a slow pulse/glow on the chevron (respects `prefers-reduced-motion`) for as long as a non-default filter/sort is applied, plus a brief one-shot toast ("Filtering: X" / "Sorted: X") when an option is picked.
- Threaded through the same URL-param -> fetch -> render pipeline as the existing tag/search/sprintId filters (`loadBoardBySlug`, `renderBoardFromData`, `updateBoardContent`, lane load-more, board refresh orchestration, router parsing).
- New i18n keys (`board.filters.assignee/allAssignees/unassigned/assignedToMe/sort/defaultOrder/newestFirst/oldestFirst/filteringOn/sortedBy`) added to all 24 locale catalogs.

## Known limitation
- Manual drag-and-drop reordering stays enabled even when a sort (newest/oldest) is active. Dragging doesn't really make sense against a non-manual sort, but disabling/hiding it felt like a separate follow-up rather than part of this change.

## Test plan
- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go test ./... -count=1 -timeout=5m`
- [x] `npm run build` (tsc, i18n verify, landing generation all pass)
- [x] New/updated test files: `board-filter-panel.test.ts` (panel open/close, option selection, URL round-trip, reload wiring for both assignee and sort), `internal/store/board_sort_test.go`, `internal/httpapi/board_sort_test.go`, `internal/mcp/board_sort_test.go` (ordering + no-dup/no-gap cursor pagination in both sort directions) — all pass
- [x] Existing filter/i18n/orchestration test suites updated for the new param and pass
- [x] Manually exercised in a local SQLite dev build: bootstrap, create project, add todos, assign one, open the filter panel, filter by assignee, sort by newest — screenshots in the PR comment thread


jordanfelle:
Screenshots from a local test run (SQLite dev build, one project with a couple of todos):

**The dropdown, open** — All assignees / Unassigned / Assigned to me / each board member:

assignee dropdown open

**Filtered to "Assigned to me"** — 1 of 4 backlog cards:

filtered to assigned to me

**Filtered to "Unassigned"** — the other 3:

filtered to unassigned

jordanfelle:
Let me know if you want this changed design wise, but this would make my life so much easier

markrai:
> Let me know if you want this changed design wise, but this would make my life so much easier

We can even add oldest/newest to this.  There are times I just want to see what were the last few items added, or the oldest item in the backlog 

As for aesthetics, what we can do is dual-purpose the existing "Search Todos" by adding a small "down notch" which expands it to these options.

This way we don't clutter the navbar - esp. on PWA 

Because filtering is a "search adjacent" action, so it should be okay.

The only thing that needs to be figured out is:

Assuming user has a filter applied/active... What indicator to give user so they know that it's currently filtering on user: "Jordan"

One option is a persistent toast at bottom:

"Filtering on: Jordan"

which goes away once user unchecks it.

Another (better) option is a slow pulse glow of this downward facing notch, and just keep a regular toast (which goes away briefly)

EDIT:

I realized there's an X to clear input at the right of the search field... So the downward notch could be to the right of that X (same color profile as the X)

jordanfelle:
Implemented both pieces of that:

**Redesigned as an expandable panel on the search box** — no more separate `<select>` in the topbar. The chevron next to "Search todos..." expands a small popover with Assignee and Sort together:

closed
panel open

**Added sort (newest/oldest)** next to assignee in the same panel — threaded through the full stack (store query ORDER BY, cursor-based lane pagination, HTTP `sort` param, MCP `board_get` tool), not just cosmetic.

**Active-filter indicator**: went with your second option — a slow pulse/glow on the chevron (persists as long as a filter/sort is active) plus a brief one-shot toast when you pick something, rather than a persistent banner:

filtered
filtered + sorted

Toast reads "Filtering: Jordan" / "Sorted: Newest first" and fades after ~2.5s; the pulse (not visible in a still frame) respects `prefers-reduced-motion`.

One known tradeoff: manual drag-and-drop reordering stays enabled even with a sort applied — dragging while sorted by newest/oldest doesn't really make sense, but disabling it felt out of scope here. Happy to follow up on that separately if you'd rather it be disabled/hidden when a sort is active.

markrai:
@jordanfelle - this is great! THANK YOU!

I'm working on the sorting quirks right now. (some of which you already foresaw) Please stand by :)

markrai (approved):
Chef's kiss!

tungstendev9 (approved):
approved.

PR #195, opened 2026-07-27
restrictive top-level default with scoped job permissions

markrai:
StepSecurity fix so the workflow level default is read-all


tungstendev9 (approved):
Do the optimizations ever end?

PR #196, opened 2026-07-27
Fix cross-lane drag disabled under chronological sort

jordanfelle:
## Summary
- 3.28.0's chronological-sort DnD guard went too far: it hid the card drag handle entirely and skipped creating any Sortable instances when `sort=newest`/`sort=oldest` was active, so cards couldn't be dragged to a *different* lane, not just reordered within one.
- DOM neighbors under chronological order are still invalid manual-rank anchors, so same-lane reordering stays disabled (`Sortable`'s `sort: false` on each lane instance). The drag handle is now always rendered, and cross-lane drops append to the end of the target lane's rank order (`afterId`/`beforeId` both omitted) instead of trying to anchor off chronological neighbors.
- Bumped to 3.28.1 with a changelog entry.

## Test plan
- [x] `npx tsc --noEmit`
- [x] `npx vitest run modules/features/drag-drop.test.ts modules/views/board-rendering.test.ts` (updated to cover the new sort:false + always-rendered-handle behavior; both pass)
- [x] `go build ./...`

### DIFF
--- a/internal/httpapi/web/modules/i18n/locales/uk.json
+++ b/internal/httpapi/web/modules/i18n/locales/uk.json
@@ -199,7 +199,7 @@
   "board.selection.multiple": "Редагувати {count} вибраних",
   "board.selection.single": "Редагувати вибране",
   "board.status.backlog": "Backlog",
-  "board.todo.dragToReorder": "Перетягни, щоб змінити порядок",
+  "board.todo.dragCard": "Перетягніть картку",
   "board.todo.moveFailed": "Не вдалося перемістити Todo",
   "board.todo.movedTo": "Todo переміщено в {lane}",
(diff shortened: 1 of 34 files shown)


tungstendev9:
yeah I would agree on this one

markrai (approved):
already reviewed

internal/httpapi/web/modules/i18n/locales/uk.json
tungstendev9, line 1: @markrai why the semantic change?

markrai: because it's mode neutral..  saying "reorder" for oldest/newest since that's not allowed

tungstendev9 (approved):
approved.

PR #197, opened 2026-07-27
smtp: use specific per-reason copy for activity update emails

jordanfelle:
## Summary
- `board.refresh_needed` activity emails all shared one generic "activity update" subject/body regardless of what actually happened (card moved, sprint closed, etc.), even though the handler already has `reason`, `actorUserId`, and the project name in hand.
- Replaces `refreshReasonCategory` with `refreshReasonInfo`, which pairs each reason with its existing category plus subject/body copy (e.g. "card moved" / "A card was moved").
- When the actor's user record is available, the body leads with their name ("Alex moved a card in Acme.") instead of the passive fallback ("A card was moved in Acme.") — same specificity as the existing assignment email, no new plumbing (`GetUser` was already on `emailNotifyStore`).
- Stays with the existing plain-text `fmt.Sprintf` style; no templates, no digest/aggregation, no i18n — matches the "easy win" scope from the discussion.

Discussion: https://github.com/markrai/scrumboy/discussions/193

## Test plan
- [x] `go build ./...`
- [x] `go test ./...`
- [x] Added `TestEmailNotifier_RefreshNeeded_CopyReflectsReasonAndActor` covering reason-specific subject/body and actor-name inclusion
- [x] Updated existing assertions in `email_notify_corrections_test.go` / `email_notify_endtoend_test.go` that checked for the old generic "activity update" copy

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 > **Upgrades:** No breaking changes for **3.7.0 ≤ v ≤ 3.28.x** unless noted below. Notable upgrade impact: **3.22.0** (MCP/OAuth), **3.24.0** (MCP tool names), **3.26.0** (MCP project tags) - see those releases.
 
+## [3.28.2] - 2026-07-27
+
+### Changed
+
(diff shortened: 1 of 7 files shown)


markrai (approved):
@jordanfelle 

The only thing I stopped short of was the case in which the user creates a Temporary board, and then shares it with an unauthenticated user, and begins to receive emails which say:

"Anonymous User" has done x activity.

I think we'll leave that alone for now, as temp boards are meant to be more "throwaway" in nature.

tungstendev9 (approved):

PR #198, opened 2026-07-28
Add per-user "cards per lane" default preference

jordanfelle:
## Summary
- Board lane page size was hardcoded to 20 in three places: the initial `GET /api/board/{slug}` fetch default, the `projects.ts` hover-prefetch, and the client-side load-more floor.
- Adds a `cardsPerLane` user preference (integer, 5-100, validated server-side alongside the existing `tagColors`/`wallpaper`/`emailNotifications` preferences) so each user can set their own default instead.
- Adds a "Cards per lane" control under Settings > Customization, wired to `PUT /api/user/preferences`.
- Frontend hydrates the preference at boot (same pattern as wallpaper/voiceflow/dashboard-sort prefs) and applies it to the initial board fetch, hover-prefetch, and the load-more floor reset.

## Test plan
- [x] `go test ./...`
- [x] `npx vitest run` (no new failures vs. main; 3 new passing tests added for the default/floor logic)
- [x] `node scripts/verify-i18n-locales.mjs` (24 catalogs pass parity, including the 5 new `settings.customization.cardsPerLane.*` keys)
- [x] `npx tsc --noEmit`


jordanfelle:
Screenshots of the new Settings > Customization control, tested locally against the built app:

**Default state (20)**


**After changing to 45 — saved and toast confirms**


Verified the value round-trips through `GET /api/user/preferences?key=cardsPerLane` after save.

markrai (dismissed):
already approved.

markrai (approved):
already reviewed.

tungstendev9 (approved):
approved. good call on the tests.
drop-down was also a good move.

PR #200, opened 2026-07-28
Add per-board option to disable sprints

jordanfelle:
## Summary
- Adds a `sprints_enabled` per-project setting (default: on) with a maintainer-only toggle in Settings > Sprints.
- When disabled, hides the Sprints tab content, the Sprint field on the New/Edit Todo dialog and Bulk Edit dialog, and the sprint filter chips on the board.
- Backend also rejects `POST /sprints` while disabled (defense in depth, not just UI hiding).
- Included in export/import (backup) round-trip; older backups without the field import as enabled.

## Test plan
- [x] `go build ./...`, `go vet ./...`, `go test ./...` (full repo)
- [x] Frontend: `tsc --noEmit`, `npm run build`, `npm test` (vitest)
- [x] Manual: bootstrapped a fresh instance, created a board, toggled sprints off/on in Settings > Sprints, confirmed the Sprint field disappears from the New Todo dialog while disabled, confirmed toggling back on restores it


tungstendev9:
I'm testing with an account which has the "Contributor" project role and I'm seeing that the Sprints tab already not visible. In other words, the maintainer would simply have to cease creating new sprints (and perhaps delete the existing one) and this would effectively mean "no sprints." 

And if all sprints are deleted (or none exist) then the associated tags (scheduled/unscheduled) also disappear:

image

The only "annoying" part which remains is an unnecessary "Sprint" dropdown for Maintainers:

{45EF380F-FD8F-4380-A245-0466B23C98E8}

If we do implement this PR, we also have to account for some MCP plumbing, as well as the rabbit hole of answering questions such as: Does disabling sprints implicitly close active sprints, and is the assignment data preserved? If re-enabled, does everything appear as before? or reset? And then the metrics aspect as well.. 

There might be a more elegant and narrower solution here. Namely, the hiding of the "Sprint" dropdown for maintainers when no sprints are being used.

What do you think @markrai ?

markrai:
> What do you think @markrai ?

Yeah those are fair points... I think we'd end up adding more complexity for little gain. Best to handle the aesthetics separately where the "Sprint" dropdown in Card Edit becomes moot if there are zero sprints available. (i.e...the system shouldn't care whether the sprints are expired or active.. otherwise we make it more rigid than necessary... (a use case might be where a person wants to add an item to represent work in a sprint which has technically ended) since everyone follows their own version of scrum.

tungstendev9 (dismissed):
approved.

markrai:
Let's hold off on this one

markrai:
@jordanfelle I have not forgotten about this! Also definitely feeling it on my home projects with my wife where the Sprint and the story points are just annoying and meaningless 😅

Will circle back to this as soon as the refactor is complete. This will also make it much safer and easier to go back to feature work.

jordanfelle:
@markrai just tell her that taking out the trash is in next sprint

markrai:
> @markrai just tell her that taking out the trash is in next sprint

Coming to this tomorrow morning as soon as the Sprint seam is done (I am midway in it) Happy Friday! ☺️

markrai:
@jordanfelle @tungstendev9 
It's late but I've done some basic testing... would be good if you guys also took a look to make sure it landed 

v3.30.0 test demployment: [ghcr.io/markrai/scrumboy:test-build ](url)

tungstendev9 (approved):
solid work. approved.

markrai (approved):
already approved.

PR #201, opened 2026-07-28
Stretch board columns to fill width when fewer than 5 lanes

jordanfelle:
## Summary
- `.board`'s CSS grid was hardcoded to `repeat(5, minmax(200px, 1fr))`, so a project's workflow with fewer than 5 columns (e.g. Backlog/In Progress/Testing/Done) left empty grid tracks and unused whitespace on the right instead of stretching the existing columns to fill the row.
- Switched to `repeat(auto-fit, minmax(200px, 1fr))`, matching the pattern already used by other responsive grids in this stylesheet.

## Before / After

Before (4 lanes, whitespace on the right):

before

After (4 lanes, fills the width):

after

Verified no regression with 6 lanes — columns still wrap/stretch evenly with no leftover whitespace.

## Test plan
- [x] `go build ./...`
- [x] `go test ./internal/httpapi/...` (web asset/static tests)
- [x] Manually verified via local run + screenshots with 4-lane and 6-lane workflows


markrai:
This is good. I think we should probably also give the user the capability to collapse into more than 1 row at some point.

image

In other words, some folks might want 7 lanes to be represented like this:

row 1:  3 lanes
row 2:  3 lanes
row 3:  1 lane.

or this:

image


i.e a rule could be: if it is an EVEN number of lanes above 5, then create a new row...  
This way we also don't break flows for folks who are already using the app. 

I think toggle is best for this. The default is that all lanes show up in the same row... but if toggle is turned on... then always attempt to break into even rows. (and then account for the odd number of lanes that are remaining... and perhaps not 'resize' them, cuz that would look janky. In other words... the invariant would be: all lane widths should be equal at any given time - even if some are "danglers."

Tangential thought to this:  Now with the 12 lane workflow before me, I'm also seeing the use-case for limiting cards below the floor of 20 we had set earlier... (I figured who wants to see less than 20 items on a lane?) BUT if someone has a 12 lane workflow... perhaps they want to see all lanes at once, and limit to 5 items each...  Just keeps Scrumboy's philosophy of being flexible without too much effort.

markrai:
The best place for the toggle should probably be under a new entry called: 

 Settings -> Customization

markrai:
@jordanfelle / @tungstendev9 
Working on this right now to expand it.

markrai (approved):
solid.

markrai (approved):
already reviewed.

tungstendev9 (approved):
approved.

PR #205, opened 2026-07-31
refactor(board): decouple reads and align MCP contracts

markrai:
Board read operations across REST and MCP transports have been unified behind a shared application service (`internal/application/board`) to keep transport wrappers thin while preserving existing permissions, pagination, and response shapes. Key contracts and behaviors are now explicitly documented and locked via contract tests, including unpaged numeric REST compatibility, MCP `board_get` tiered validation precedence, sprint ID identity (stored row IDs vs. REST project-local numbers), canonical slug identity returns, sanitized JSON-RPC structured error handling, and approved legacy metadata exposure. Additionally, temporary board reads in MCP now treat activity maintenance updates as best-effort to prevent read failures on throttled lifetime refreshes, while ensuring pagination maps stay consistently aligned across tool catalogs and JSON-RPC payloads.

see CHANGELOG.md for the full details.


tungstendev9 (approved):
Approved. I believe the next item should be write-side orchestration like todo command handling.

also as you mentioned already the MCP mutations don't emit a board.refresh_needed. This leaves open browser clients stale until another refresh is done. Not a priority at this point.

PR #206, opened 2026-07-31
use absolute overlay on desktop for .no-results

markrai:
On desktop/tablet, .no-results is now position: absolute over .board, so it’s out of the grid entirely and lanes stay at their normal widths. Mobile (≤620px) is unchanged.


tungstendev9 (approved):
thank you. approved.

PR #207, opened 2026-08-01
Maintenance/refactor write side

markrai:


tungstendev9:
I will look at this tomorrow morning. 

Next step is to compare todo create and todo delete. Delete is probably the smaller boundary and Create likely has more maintainability value but probably more assignment, positioning, refresh, and notification items which will come up

tungstendev9 (approved):
approved.

PR #209, opened 2026-08-02
Refactor/todo creation

markrai:
Route REST and MCP todo creates through a shared application service: Same write-side pattern as update/move: contract-test first, then a transport-neutral create boundary, prepared REST/MCP services, and thin adapters - behavior preserved.


tungstendev9 (changes requested):
add the new ver and changelog please

tungstendev9 (approved):
approved

PR #211, opened 2026-08-02
Refactor/workflow column mutation

markrai:
REST and MCP workflow col create, update, and delete now share command boundaries instead of owning write logic in the transports. Part of an on-going refactor effort.


tungstendev9 (approved):
approved

PR #212, opened 2026-08-03
Refactor/project membership

markrai:
REST/MCP member add, role update, and remove now go through internal/application/membership, with transports left as thin wrappers and contract tests locking the paths.


tungstendev9 (approved):
might want to clean up the inline comments at some point

PR #213, opened 2026-08-04
security: pin postcss and undici for npm advisories

markrai:
Override transitive PostCSS/Undici to patched releases so Vite and jsdom no longer resolve the six reported GHSA advisories.


markrai:
@tungstendev9 yes, I know it's late but patching needs to happen promptly :)

tungstendev9 (approved):
approved. good quick catch.

PR #216, opened 2026-08-05
Add customizable per-board priority tiers for todos

jordanfelle:
## Summary
- Adds a `priority` field to todos, modeled as per-project customizable tiers (add/rename/recolor/delete) rather than a fixed global enum, mirroring the existing workflow-column feature. New projects seed with four default tiers: Low, Medium, High, Urgent.
- New Settings → Priorities tab (maintainer role required) manages tiers, mirroring the Workflow tab's UX. The todo dialog's Priority select is now populated dynamically from the project's tiers, and the selected tier renders as a colored badge on board cards.
- New REST (`GET/POST /api/board/{slug}/priorities`, `PATCH`/`DELETE .../priorities/{key}`, `GET .../priorities/counts`) and MCP (`priorities_list/create/update/delete`) endpoints, following the same `internal/application/{workflow,priority}` mutation-service pattern the workflow-column feature was recently migrated to. `todos_create`/`todos_update` (REST + MCP) gained a `priorityKey` field.
- Deleting a tier is blocked while any todo still references it, or if it would leave the project with zero tiers.
- Backup export/import round-trips both custom tiers and each todo's priority.
- i18n: added `settings.priorities.*`, `todo.fields.priority`, and tooltip keys across all 24 locale catalogs. The ~23 non-English translations are my own best-effort pass — flagging for a native-speaker spot check before merge, especially for the less common locales (sw, ur, fa, ms, bn, th, vi).

**Deliberately out of scope for this PR** (called out in the CHANGELOG): filtering/sorting the board by priority. The board read/query pipeline doesn't support it yet on either transport; only per-tier "in use" counts are exposed (for the Settings tab's delete guard). Happy to follow up in a separate PR if that's wanted.

## Test plan
- [x] `go build ./...`, `go vet ./...`, `go test ./...` all green (migrations, store CRUD + validation, application-layer REST/MCP mutation services, REST + MCP transport contract tests for create/update/delete/counts, failure-silence/refresh-event assertions)
- [x] `npm run build` (tsc + i18n parity + landing-page verify) and `npm test` for the touched frontend files (`settings-priorities`, `todo-submit`, `board-rendering`) all green
- [x] Manually exercised via a local run: added/renamed/recolored/deleted tiers in Settings → Priorities, confirmed delete is blocked while a tier is in use or would leave zero tiers; set a todo's priority via the dialog and confirmed the card badge renders with the tier's color; confirmed export/import round-trips custom tiers and per-todo priority
- [ ] Native-speaker review of the non-English i18n strings added in this PR


tungstendev9:
@markrai  The priority sort would be lovely. I could handle the planning around that next
@jordanfelle Thank you so much for doing this!

markrai (approved):
already reviewed

tungstendev9 (approved):
approved.

PR #225, opened 2026-08-11
Add columnKey filter to board_get MCP tool

jordanfelle:
## Summary

`board_get` always queries and returns every workflow column (backlog/doing/done/...) even when a caller only wants one. On a board with hundreds or thousands of todos, listing a single lane still means paginating through every other lane's `ListTodosForBoardLane`/`CountTodosForBoardLane` calls and serializing them into the response — this doesn't scale as board size grows.

The per-lane store queries already scope to a single `columnKey` (`WHERE column_key = ?`); `board_get`'s MCP schema just never exposed a way to ask for one lane only. The app's own board UI already has this via `GET /api/board/{slug}/lanes/{status}` for infinite-scroll pagination — this PR gives the MCP tool the same capability.

- Add an optional `columnKey` field to `board_get`'s input schema.
- Validate it against the project's own workflow columns (same pattern as the existing `sprintId`/`cursorByColumn` validation) and return a `columnKey` validation error for an unknown key.
- When set, skip the list/count store calls entirely for every other column instead of querying and discarding them.
- Omitting `columnKey` preserves the existing all-columns behavior exactly.

## Test plan

- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `golangci-lint run ./internal/mcp/... ./internal/application/board/...` — no new findings in touched files
- [x] `go test ./...` — full suite green
- [x] Added unit coverage in `internal/application/board/mcp_read_test.go` (matching-key scoping, whitespace trimming, unknown-key validation, empty-key backward compatibility)
- [x] Added contract coverage in `internal/mcp/board_get_validation_contract_test.go` (unknown key fails before any lane query; matching key scopes list/count calls to one column)
- [x] Added an end-to-end HTTP+MCP test in `internal/mcp/board_get_projection_integration_test.go` confirming the response contains only the requested column and `totalCountByColumn` is scoped to it


tungstendev9 (approved):
approved.

PR #227, opened 2026-08-12
Use yellow for the default Low priority tier color

### LINKED ISSUE #226: UI: priority pill/icon on cards
Image

We added the priority designation per todo, but we need to also reveal it now...  and perhaps colorize based on how important it is:

red - urgent
orange - high 
amber - medium
yellow - low

I think this color scheme would be a good fit.. (tested in both light and dark mode)

@jordanfelle / @tungstendev9  - If you wanna pick it up sure.. or I can go ahead :) no pressure.

Also... need to ensure that mobile view renders this correctly as well.

jordanfelle:
## Summary
- Closes most of #226 (UI: priority pill/icon on cards) — the pill is already rendered on cards with per-tier colors (added in #216), and medium/high/urgent already use amber/orange/red as requested there. This just fixes the last mismatch: the seeded default `low` tier was still generic gray (`#9CA3AF`) instead of yellow.
- New default: `low` → `#EAB308` (yellow), matching the red/orange/amber/yellow scale from the issue.
- Confirmed the pill is a plain non-interactive `<span>` (matches the "shouldn't be clickable" conclusion in the issue thread) and that mobile card layout (`.card__badges` grid placement) already handles the badge without overflow — no CSS changes needed there.
- Only affects the *default* seed for newly-created projects; existing projects that already have a `low` tier keep whatever color they have (including any prior custom edits), since project owners can always recolor tiers in Settings → Priorities.

## Test plan
- [x] `go build ./...`
- [x] `go test ./... -count=1 -timeout=5m` (full suite green)
- [x] Confirmed no test hardcodes the previous default `#9CA3AF` for a freshly-seeded project (the one match in `priorities_concurrency_test.go` is an unrelated import-fixture color)

Closes #226

### DIFF
--- a/internal/store/priorities.go
+++ b/internal/store/priorities.go
@@ -76,7 +76,7 @@ func priorityTierCountReason(count int) string {
 
 func defaultPriorityTiers() []PriorityTier {
 	return []PriorityTier{
-		{Key: "low", Name: "Low", Color: "#9CA3AF", Position: 0},
+		{Key: "low", Name: "Low", Color: "#EAB308", Position: 0},
 		{Key: "medium", Name: "Medium", Color: "#F59E0B", Position: 1},
 		{Key: "high", Name: "High", Color: "#F97316", Position: 2},
 		{Key: "urgent", Name: "Urgent", Color: "#EF4444", Position: 3},


markrai:
@jordanfelle oh whoops! my bad! I added the designations on one instance, but tested on an older one 😅

No, actually I like your initial implement much better! because if a priority is "low" then the gray makes sense, because you don't really want it to "catch your eye" and the gray implies that it's well... "low priority"

So I think we'll leave it as it is :)

PR #228, opened 2026-08-12
Add priority filtering to the board filter dropdown

jordanfelle:
## Summary

Adds the ability to filter the board by priority tier, mirroring the existing Assignee/Sort filters end to end.

- **Backend**: new `priority` query param, parsed the same way as `assignee` (`""` = all, `"none"` = no priority set, a tier `key` = that specific tier), threaded through the store SQL builders, REST board/lane read handlers, and the `board_get` MCP tool.
- **Frontend**: a new "Priority" section in the existing filter dropdown popover — "All priorities", "No priority", and one button per project priority tier (tinted with the tier's color), matching the Assignee section's markup/interaction pattern. The `priority` param is threaded through every board reload/invalidate path (drag-drop, bulk-edit, settings dialogs, foreground-resume, realtime SSE, initial route load) so the filter survives all the same flows assignee/sort already do.
- i18n keys added to all 24 locale catalogs (`board.filters.priority`, `allPriorities`, `noPriority`); non-English translations are straightforward short UI phrases but flagging for a native-speaker spot check before merge, per this project's convention.

## Test plan
- [x] `go build ./...`
- [x] `go test ./... -count=1 -timeout=5m` — full suite green
- [x] `npm run verify:i18n` — all 24 catalogs pass parity
- [x] `npx tsc --noEmit` — zero type errors
- [x] `npx vitest run` — 796/797 passing; the sole failure (`keybindings-i18n.test.ts`, a `localStorage` API quirk under this sandbox's Node version) is unrelated to this change and reproduces identically against unmodified `main`
- [x] Manual security review of the new SQL builder, HTTP/MCP query parsing, and frontend HTML rendering — no findings; the whole feature is a parameterized-query, escaped-HTML mirror of the existing assignee filter


markrai (approved):
making a version bump

tungstendev9 (approved):
approved.

PR #230, opened 2026-08-13
Patch/pr 228

markrai:
Fixes three issues introduced with priority filtering: resolves the collision between the `none` priority-tier key and the special “No priority” filter by moving the sentinel to an unambiguous value; automatically clears stale priority filters when a selected tier is deleted so users aren’t left with an empty board; and adds the missing `priority` parameter to the MCP `board_get` discovery schema so MCP clients can correctly discover and use priority filtering.


tungstendev9 (approved):
approved. good catch!

PR #233, opened 2026-08-14
Refactor/legacy todo

markrai:
Work completed:

Refactored the legacy numeric To Do mutation paths to use the application service layer, completing the migration of update, move, and delete operations away from transport-owned business logic. The refactor preserves the existing legacy API contracts, including global numeric ID semantics, authorization and validation precedence, event and refresh behavior, projection handling, and anonymous/temporary board behavior, while consolidating mutation orchestration behind prepared services. Characterization and service-level regression tests lock down the legacy behavior throughout the migration, reducing duplication and bringing the legacy endpoints in line with Scrumboy’s modern read/write architecture without intentionally changing externally observable behavior.

Deferred items:

@tungstendev9 Numeric MOVE currently permits mutation on an expired temporary board. Phase 22 deliberately preserved this pre-existing behavior. It should be addressed separately as an explicitly behavior-changing correctness/security task with dedicated policy and compatibility review


tungstendev9 (approved):
ugh you again 😁

approved.

PR #234, opened 2026-08-14
Track who opened each todo, notify them on updates

jordanfelle:
## Summary
- Adds a nullable `created_by_user_id` column to todos, set from the authenticated actor at `CreateTodo` time (NULL for anonymous/legacy todos, same convention as `projects.creator_user_id`). Threaded through board queries, backup export/import/restore, project duplication, the REST and MCP DTOs, and a new "Opened by <name>" line in the todo dialog.
- Adds a `createdByMe` email-notification preference (defaults on, mirrors `assigned`) and a new `todo.creator_notified` event, fired from `UpdateTodo`/`MoveTodo` whenever the acting user differs from the todo's creator. Delivered as an email via the existing notifier pipeline, and as an in-app toast pushed only to the creator's own SSE user-channel.

## Test plan
- [x] `go build ./...`, `go vet ./...`, `go test ./... -count=1` all pass
- [x] Frontend `tsc --noEmit` and `npm run verify:i18n` (24 locale catalogs) pass
- [x] New Go tests: `EmailNotifyPref` parse/round-trip for `createdByMe`, `emailNotifier` opt-in/opt-out email tests, and a full eventbus regression test asserting `todo.creator_notified` reaches the creator's SSE user-channel when someone else edits their card
- [x] Manually traced the "Opened by" UI path (todo dialog reads `createdByUserId`, resolves the name via existing board-member lookup, hidden when unknown/legacy)


tungstendev9:
a few issues:

1. removed creator still receive project data:

`created_by_user_id` is historical state but the PR treats it like current authz.  so they will still get SSE.

2. There's an ID-reuse edge case hiding in here... in terms of backup and restore of creator ID... 
probably better to  make backup/copy creator attribution depend on current project membership. Historical authoriship and project membership should remain separate concepts.

i.e. if Bob creates a todo and later leaves the project, it is still true as historical fact that bob created it... so

CreatedByUserID should be used to describe historical author, but never grant access or notification eligibility.

3. the notification is implemented below mutation-service boundary. 

Actually @markrai - do you want to see first whether some architectural changes would make sense first PRIOR to us handling this PR? 

A. Instead of store publishing creator notificatons itself, UpdateTodo/MoveTodo should return enough info for the app service to decide . See AssignmentChanged - we are already beginning to do this.

B. Introduce applicatoin level notification/effects coordinator. The prepared mutation services should own "What external effects follows this successful mutation?" and then coordinator decides the rest. Don't bury those rules in SSE and email handlers.

C. Formalize historical identity concept... 

We have to think this through and make the changes small since we're already in the middle of a refactor, but moves would prevent future headaches and paradoxes 😅

markrai:
@tungstendev9 I took a look and I don't think we need a scaffolding or prepping prior to bringing in and adapting PR 234.

Will take another look tonight or tomorrow morning to be sure.

markrai:
@tungstendev9 So far so good. I'm going to work on the authorized creator delivery tomorrow morning, and then creator email delivery after that.

tungstendev9 (approved):
approved

PR #237, opened 2026-08-16
fix post-mutation board refresh

markrai:
fixes a post-mutation board refresh bug that could leave filtered boards showing stale data after editing, creating, moving, or deleting todos. The manual refresh path was not preserving the active assignee, sort, and priority context, causing otherwise valid board responses to be rejected by the existing request-identity guard. The fix passes the full board filter/sort context through post-mutation reloads, aligning them with the rest of the refresh infrastructure, and adds regression coverage for filtered refresh behavior, realtime cancellation interaction, and rendering of updated lane and priority state.


tungstendev9 (approved):
approved.

PR #238, opened 2026-08-16
ci: increase Go test timeout

markrai:


tungstendev9 (approved):
good call. our CI needs a performance review at some point.

PR #240, opened 2026-08-17
fix(board): persist sort preference and dedupe assignee filter

markrai:
- Your own name is now filtered out of the Assignee section (Assigned to me was always there, thus redundant)
- Your sort preference for Newest First, Oldest First, etc. is retained.


tungstendev9 (dismissed):
ah yes. much needed. Approved.

tungstendev9 (approved):
approved.

PR #241, opened 2026-08-18
ci: verify committed dist matches frontend sources

1269:
## Summary

The frontend is served from the committed `internal/httpapi/web/dist` directory, which is embedded into the release binary via `go:embed` (`internal/httpapi/server.go`) and shipped as-is by the Docker build. CI never rebuilt it, so two failure modes could pass CI and reach a release:

1. **Stale dist** — a contributor edits `modules/*.ts` but forgets `npm run build`; CI stays green while the shipped app runs old JS.
2. **Type errors** — the `npm test` script does not run `tsc`, so TypeScript errors in `modules/` were never checked on PRs.

## Changes

- **New CI steps** in the frontend job: clean rebuild (`rm -rf dist && npm run build`) followed by `git diff --exit-code` / `git status --porcelain` on `dist`. The clean rebuild means orphaned files show up as deletions and fail the check too. Since `npm run build` runs `tsc`, every PR is now type-checked as a side effect.
- **Removed 8 orphaned files** under `dist/dialogs/` (`api.js`, `theme.js`, `types.js`, `utils.js`, `dom/`, `state/`) left behind by an earlier source refactor. Nothing references them and a clean rebuild does not recreate them — they were dead weight embedded in every release binary.

## Verification

- Ran a clean rebuild locally: output is byte-identical to committed `dist` (the `copy:i18n` and `generate:landing` steps are deterministic; verified via `diff -rq`).
- The rebuild does not recreate the deleted orphans, and `grep` finds no imports or references to them in `dist`, `modules`, `index.html`, or `app.js`.
- The new verify steps pass locally against this branch.


markrai:
@1269  Oh thank you! We've been doing this manually, and not gonna lie - there's been at least one lapse (happened this weekend, in fact!)

tungstendev9:
yassssss. I'm a bit embarrassed we weren't doing this already 😅

tungstendev9 (approved):
approved.

PR #242, opened 2026-08-19
Security/go 1.25.13

markrai:
**Go 1.25.13 toolchain** - Bump the `go.mod` toolchain from `go1.25.12` to
  `go1.25.13` and pin the Docker build image to `golang:1.25.13-alpine` by
  multi-platform manifest digest so the standard-library vulnerabilities
  fixed in that patch are no longer reported by OSV Scanner. The `go 1.25.0`
  language-version directive is unchanged.


tungstendev9 (approved):
I will allow it. approved.

PR #255, opened 2026-08-22
Refactor/tag mutations

markrai:
all four mutation families independently prove adapter -> internal/application/tag -> store ownership, the forbidden direct-handler searches are clean, REST publication and MCP silence are structurally verified, characterization tests remain intact, and the full repository/static/build/docs gates pass.


tungstendev9 (approved):
on a weekend man??? 😁
approved.

PR #263, opened 2026-08-25
Anonymous mode: selective hiding of controls which do not apply in Settings

markrai:
Removed items when running in anonymous mode from settings, which are not applicable/noise.


tungstendev9 (approved):
finally! I've been on your back about this for weeks now! 😅

PR #265, opened 2026-08-26
macOS executables github workflow

markrai:
produces executables for macOS (both amd64 & arm64)


tungstendev9 (approved):
i see the test branch succeeded too, so that is promising. 
approved.

PR #272, opened 2026-09-06
fix: prevent SSRF in webhook delivery

markrai:
## Summary

Hardens outbound webhook delivery against server-side request forgery (SSRF).

Webhook destinations are now resolved and classified at delivery time, and connections are made directly to vetted IP addresses rather than allowing Go's default HTTP transport to independently resolve the hostname.

## Changes

* Add shared outbound IP classification and vetted-IP dialing in `internal/safehttp`
* Block loopback, private, link-local, metadata, IPv6 ULA, multicast, unspecified, and other special-purpose address ranges
* Prevent DNS rebinding by dialing the validated IP directly
* Reject redirects during webhook delivery
* Disable environment `HTTP_PROXY` / `HTTPS_PROXY` use for webhooks
* Preserve HTTP/2 support with the hardened transport
* Treat forbidden destinations as permanent delivery failures
* Apply delivery-time protection to webhook URLs already stored before upgrade
* Add fail-fast validation for literal unsafe webhook destinations at creation time
* Reuse the shared address-classification logic for Agenda calendar fetches
* Preserve public `http` and `https` webhook support and existing webhook HMAC/body semantics

## Compatibility

Existing webhooks targeting loopback, LAN/private, Docker/internal, link-local, metadata, or other blocked destinations will no longer be delivered.

Webhook delivery also no longer honors environment `HTTP_PROXY` / `HTTPS_PROXY` settings.

Publicly routable `http` and `https` webhook endpoints remain supported.

## Testing

Added regression coverage for:

* direct loopback and private destinations
* IPv4 and IPv6 unsafe ranges
* IPv4-mapped IPv6 addresses
* mixed public/private DNS responses
* DNS rebinding / second-resolution prevention
* redirects to private destinations
* already-stored unsafe webhook URLs
* proxy and transport configuration
* retry classification
* HTTP/2 preservation
* webhook body, headers, and HMAC behavior

Focused test suites for `safehttp`, `store`, `httpapi`, and calendar pass, along with `git diff --check`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,18 @@
 # Changelog
 
-> **Upgrades:** No breaking changes for **3.7.0 ≤ v ≤ 3.33.x** unless noted below. Notable upgrade impact: **3.22.0** (MCP/OAuth), **3.24.0** (MCP tool names), **3.26.0** (MCP project tags), **3.29.0** (MCP JSON-RPC error/`board_get` identity), **3.30.0** (reversible per-project sprint capability), **3.31.0** (per-project priority tiers), **3.33.0** (Agenda ICS feeds need `SCRUMBOY_ENCRYPTION_KEY`) - see those releases.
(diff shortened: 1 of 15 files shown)


tungstendev9:
that was a bit random!

markrai:
> that was a bit random!

ha! not that random... it's an advisory fix

tungstendev9 (approved):
approved.

PR #281, opened 2026-09-13
Restrict SSO auto-signup to specific email domains

jordanfelle:
## Summary
- Adds `SCRUMBOY_OIDC_ALLOWED_EMAIL_DOMAINS` (comma-separated, case-insensitive) so operators can restrict which email domains may auto-provision a **new** account on first SSO login. Unset/empty = unrestricted (current behavior, unchanged).
- Existing users are never affected — the check only runs on the new-identity signup path in `handleOIDCCallback`, never on repeat logins.
- The very first user (instance bootstrap, before any user exists) is always exempt, so a restrictive list configured on day one can't lock an operator out of their own fresh instance — especially combined with `SCRUMBOY_OIDC_LOCAL_AUTH_DISABLED=true`.
- Linking an OIDC identity to an already-authenticated existing account (`oidc.FlowLink`, Settings → Connect SSO) is intentionally out of scope for this check — documented in `docs/oidc.md`.
- New `auth.oidc.error.domain_not_allowed` toast, added to `en.json` bootstrap catalog and translated into all 22 other locale files (not left as English placeholders).

## Adversarial review
Ran a hostile adversarial-review pass against the diff before opening this PR. Findings and resolution:
- **CONFIRMED** — `oidc.FlowLink` (linking SSO to an existing account) bypassed the allowlist with no docs callout → documented explicitly as out of scope (linking is not signup).
- **CONFIRMED** — first-user bootstrap could be permanently locked out by a restrictive list set before any user exists (especially with local auth disabled) → fixed with an explicit `CountUsers() == 0` exemption, plus a new test (`TestOIDCFirstUserBootstrapExemptFromDomainRestriction`) and a docs callout.
- **CONFIRMED** — the two new integration tests only checked the redirect URL, never that no user/row was actually created → both tests now assert against the store (`GetUserByOIDCIdentity` still `ErrNotFound` for the rejected case; user exists for the allowed case), and the test helper was extended to expose the store for this.
- **CONFIRMED** — all 22 non-English locale files got the raw English string copied in instead of a real translation (inconsistent with every other key in those files) → replaced with actual translations in each locale; `pseudo.json` correctly kept its `[!! ... !!]` bracket convention.
- Verified clean: domain-matching logic (case-insensitivity, no subdomain-escalation gap, multiple `@` handling), check ordering relative to DB writes, no info leak in the redirect.

## Test plan
- [x] `go build ./...`, `go vet ./...` clean
- [x] `go test ./...` — full suite green
- [x] `gofmt -l` clean on touched files
- [x] `golangci-lint run` — only pre-existing issues in untouched files
- [x] `tsc --noEmit` clean
- [x] `npm run verify:i18n` — locale key parity across all 24 catalogs passes
- [ ] Frontend `vitest` suite: could not get a working run in my sandbox (missing `jsdom` peer breaks the DOM test environment repo-wide, confirmed identical failure count against unmodified `main` — pre-existing environment issue, not something this PR introduces or fixes)


markrai (dismissed):
nice! thank you!

tungstendev9 (approved):
ugh! i18n. approved.

PR #283, opened 2026-09-13
fix header formatting on README.md

markrai:
revert header formatting


tungstendev9 (approved):
Do we really need an "approval" for this? 😄

PR #286, opened 2026-09-13
ci: install Capacitor dependencies in release workflows

markrai:


tungstendev9 (approved):
is today continuous integration day? 🤔
approved. 😁

PR #287, opened 2026-09-13
Ci/augment MacOS release workflow

markrai:
This effort fixed and hardened Scrumboy’s executable release workflows after the macOS v3.34.0 build failed because frontend tests imported Capacitor modules without installing `mobile/capacitor` dependencies. The macOS and Windows workflows were aligned with normal CI by installing and caching both frontend and Capacitor dependencies, restoring successful macOS builds and preventing the same failure from reaching Windows later. The macOS workflow was also extended with a safe manual release-repair mode that can rebuild missing assets for an existing tag, validate and peel annotated tags to the exact immutable commit SHA, verify the checkout before building, preserve the normal attestation/provenance pipeline, and replace only matching macOS release assets without affecting unrelated binaries or releases.


tungstendev9 (dismissed):
approved

tungstendev9 (approved):
approved.

